PR #19, opened 2019-10-08
Add Ctrl-r to cycle repeat mode; fix duplicate albums

### LINKED ISSUE #17: Duplicate Albums
Some of the albums of an artist are pushed twice. Using Linux Mint 19.1 Tessa.

duplicate

baxtea:
Included a keybind for repeat modes, which cycles through Off -> All -> Track -> Off as in the Electron client. Interface has a small visual update to show the repeat status next to the shuffle status, seen [here](https://i.imgur.com/MYKgbRN.png). The keybind inherits from shuffle (both ctrl+first letter) which seems consistent to me, but feel free to change it if you see fit.

Also closes #17, which was caused by the client not choosing between different versions of the same album available to disjoint sets of countries. It's not a permanent fix though, as it just follows the example set in `input.rs` by always requesting the UK versions. It is a little more future-proof in that both now make requests according to a new `App::country` field, but it's still essentially hard-coded. New search results for queen seen [here](https://imgur.com/HC5Bb7X.png).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add `Ctrl-r` to cycle repeat mode
+
 ## [0.0.2] - 2019-09-17
 
 ### Added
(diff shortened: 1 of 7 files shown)


Rigellute:
@all-contributors please add @baxtea for code

PR #27, opened 2019-10-08
Get a new token when the token expires

fangyi-zhou:
Not sure how to test this change, advice appreciated

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,10 +7,11 @@ mod util;
 
 use clap::App as ClapApp;
 use rspotify::spotify::client::Spotify;
-use rspotify::spotify::oauth2::{SpotifyClientCredentials, SpotifyOAuth};
+use rspotify::spotify::oauth2::{SpotifyClientCredentials, SpotifyOAuth, TokenInfo};
 use rspotify::spotify::util::get_token;
 use std::cmp::min;
 use std::io::{self, Write};
+use std::time::{Duration, Instant};
 use termion::cursor::Goto;
 use termion::event::Key;
 use termion::input::MouseTerminal;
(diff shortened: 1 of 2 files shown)


Rigellute:
I'm going to leave the app open for 1 hour, and then try to play a track 😄 .

Hilarious way to test.

Will get back to you.

src/main.rs
```diff
@@ -142,6 +153,18 @@ fn main() -> Result<(), failure::Error> {
...
 
+                if Instant::now() > token_expiry {
+                    // refresh token
+                    if let Some(new_token_info) = get_token(&mut oauth) {
+                        let (new_spotify, new_token_expiry) = get_spotify(new_token_info);
+                        spotify = new_spotify;
+                        token_expiry = new_token_expiry;
+                        app.spotify = Some(spotify);
+                    } else {
+                        // TODO: Handle error
```
Rigellute: For now, on error let's just exit like this

```rust
println!("\nFailed to refresh authentication");
break;
```

The new line in the print is so the log will show in the console.

Rigellute:
I've been testing your fork and it's looking good I think!

If you add the `break` when auth fails, this can be merged 👍

fangyi-zhou:
Rebased and added changelog

stevensonmt:
I'm curious how spotify-tui works with tokens. I am using it with spotifyd, and I thought spotifyd was requesting the API token. My solution to the token expiring was to have spotifyd.service restart every 3500 seconds. Does this PR alleviate the need for restarting the spotifyd.service?

Rigellute:
Yes, this PR fixes the auth token expiring issue.

`spotify-tui` uses the Spotify Web API to control playback on devices (such as spotifyd), and so needs it's own auth separate to spotifyd.

You can think of `spotify-tui` as a "controller" for spotify devices.

Issue #26, opened 2019-10-08
Fatal exit when type not english characters


PxyUp:
When you try to find any not English characters application did fatal exit

Rigellute:
Huh interesting, could you share the characters you used that caused the fatal exit?

PxyUp:
@Rigellute 'ыаываываайцайцайцайцайцаймямцу'

fangyi-zhou:
This is caused by inserting into a string outside word boundary, I'm submitting a PR to fix it

filiphe:
Here's a test that currently fails on "non-english" input.

```   
#[test]
    fn test_input_handler_on_enter_non_english_text() {
        let mut app = App::new();

        app.input = "My tex".to_string();
        app.input_cursor_position = app.input.len().try_into().unwrap();

        handler(Key::Char('ö'), &mut app);
        handler(Key::Char('ä'), &mut app);
        handler(Key::Char('å'), &mut app);
        handler(Key::Char('á'), &mut app);
        handler(Key::Char('þ'), &mut app);


        assert_eq!(app.input, "My texöäåáþ".to_string());
    }
```
Result from `cargo test`
```
---- handlers::input::tests::test_input_handler_on_enter_non_english_text stdout ----
thread 'handlers::input::tests::test_input_handler_on_enter_non_english_text' panicked at 'assertion failed: self.is_char_boundary(idx)', src/liballoc/string.rs:1287:9
```

PR #30, opened 2019-10-08
Insert input into correct character boundary (#26)

fangyi-zhou:
Fatal crash is due to
String::insert(https://doc.rust-lang.org/std/string/struct.String.html#method.insert)

The document says ""Panics if idx is larger than the String's length, or
if it does not lie on a char boundary.""

The solution is to find the correct index to insert into. Alternatively,
we have use a Vec<char> instead of String, but that requires more effort
to change.


fangyi-zhou:
image
The input cursor is slightly off, but I guess it's better than crashing

Rigellute:
Thanks for this again! Really appreciate your contributions.

I wonder if this cursor positioning issue could be solved using this library https://github.com/unicode-rs/unicode-width

So we can get the actual width of the characters and position the cursor properly 🤔

fangyi-zhou:
I'll try that as well

fangyi-zhou:
I discovered that width of the character doesn't always equal to the size of character...

src/handlers/input.rs
```diff
@@ -94,14 +108,23 @@ pub fn handler(key: Key, app: &mut App) {
...
+            app.input.insert(insert_idx, c);
+            app.input_idx += 1;
+            let width: u16 = UnicodeWidthChar::width(c).unwrap().try_into().unwrap();
+            app.input_cursor_position += width;
         }
         Key::Backspace => {
             if !app.input.is_empty() {
-                app.input.pop();
-                app.input_cursor_position -= 1;
+                // FIXME: Backspace always remove the last character
```
fangyi-zhou: Filed issue #33

Rigellute:
Wow this looks tricky. I'll put some thought into this based on what you've done here.

Rigellute:
This is looking good.

Would you be able to add some non english char tests to cover insert, backspace?

Something like this

```rust
    #[test]
    fn test_input_handler_on_enter_text_non_english_char() {
        let mut app = App::new();

        app.input = "ыа".to_string();
        app.input_cursor_position = app.input.len().try_into().unwrap();
        app.input_idx = app.input.len();

        handler(Key::Char('ы'), &mut app);

        assert_eq!(app.input, "ыаы".to_string());
    }
```

And if you have time, test that the `Left` press doesn't go out of bounds?

Thanks again for this amazing work!

fangyi-zhou:
Sorry for the delay, I've rebased and added tests

PR #60, opened 2019-10-13
small tweak to volume controls

stuarth:
allow volume to be lowered more naturally to 0


Rigellute:
Thanks for doing this, but for now I prefer the previous implementation due to avoiding making the API request when the volume is at either extreme.

Please feel free to apply your Rust knowledge to improve other parts of this app!

PR #67, opened 2019-10-15
Add artists view

svenvNL:
This adds the basic implementation for an artists view. I have not added pagination, in my opinion it is better to do this in a future commit or PR.

I have chosen to present the result in a table and not in a list because I would like to show the saved songs by the user per artist like the Spotify desktop application does. When an artist is selected, the user will be shown all albums of an artist, this should probably be filtered in another commit or PR.

screenshot_2019-10-15-150745

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,6 @@ The goal is to eventually implement almost every Spotify feature.
 - See if a song is liked
 - Be able to scroll through result pages in every view
 - View Library "Made for you"
-- View Library "Artists"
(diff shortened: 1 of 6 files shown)


svenvNL:
@Rigellute I have searched trough the API documentation tonight, but was not able to find an API that gives back the amount of saved songs per artist. I will do a little more investigation tomorrow.

Rigellute:
Hmm... I can't see how to do this with a public API.

Anyway, merging this now.

Great work 👍

PR #68, opened 2019-10-15
Get country from user profile

svenvNL:
Country is used by Spotify to determine which songs can be played by an user. By using the country from the users profile, the user does not have to configure it by themselves and the user will not be surprised when a song is not able to play because they have configured another country then in there Spotify account settings.


svenvNL:
@Rigellute Thanks for the feedback! I will change the code tomorrow.

svenvNL:
@Rigellute wasn't aware Rust has this syntax available. I have changed the code.

Rigellute:
Great work @svenvNL!

Rigellute:
@all-contributors please add @svenvNL for code

PR #73, opened 2019-10-16
Handle delete key

synth-ruiner:
dunno that anyone else is clamouring for this, but why not, right? :)


fangyi-zhou:
I think this needs to be rebased on #30 now

Rigellute:
Ah yes, good point @fangyi-zhou.

synth-ruiner:
Thanks- I noticed the rustfmt issue but haven't had time to add a line break! I'll rebase and fix it tonight.

synth-ruiner:
I've tried to follow the new format, and added a test for deleting a multibyte unicode character.

funnily enough I can't actually test this at home since I don't use spotify personally! sorry :)

Rigellute:
Looks like the commits have gone awry. 

Not sure what's gone wrong here but guessing this could be because of using `merge` rather than `rebase`? 

If things are too messed up, might be worth creating a new PR?

synth-ruiner:
I picked rebase in my IDE but maybe it had other ideas. I'll see if I can sort it out

PR #74, opened 2019-10-17
Show if song is liked in the playbar

svenvNL:
I have chosen to put the text [liked] after the song title if the song is liked and nothing if not. I'm not sure this is the best solution, but was not able to think of something better.

Screenshot from 2019-10-17 15-05-28

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -199,6 +199,7 @@ pub struct App {
     pub saved_album_tracks_index: usize,
     pub api_error: String,
     pub current_playback_context: Option<FullPlayingContext>,
+    pub current_playback_is_liked: Option<bool>,
     pub devices: Option<DevicePayload>,
     pub input: String,
     pub input_cursor_position: u16,
@@ -248,6 +249,7 @@ impl App {
             small_search_limit: 4,
             api_error: String::new(),
             current_playback_context: None,
+            current_playback_is_liked: None,
             devices: None,
             input: String::new(),
             input_cursor_position: 0,
@@ -298,8 +300,14 @@ impl App {
             let context = spotify.current_playback(None);
             if let Ok(ctx) = context {
                 if let Some(c) = ctx {
-                    self.current_playback_context = Some(c);
+                    self.current_playback_context = Some(c.to_owned());
                     self.instant_since_last_current_playback_poll = Instant::now();
+
+                    if let Some(track) = c.item {
+                        if let Some(id) = track.id {
+                            self.get_song_is_liked(id);
+                        }
+                    }
                 }
             };
         }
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -735,4 +743,18 @@ impl App {
             };
         };
     }
+
+    pub fn get_song_is_liked(&mut self, id: String) {
```
Rigellute: Could rename this function to `current_user_saved_tracks_contains` and let it accept an vec of ids.

```rust
pub fn current_user_saved_tracks_contains(&mut self, ids: Vec<String>) {
```

src/ui/mod.rs
```diff
@@ -559,6 +559,16 @@ where
                 RepeatState::Context => "All",
             };
 
+            let liked_text = if let Some(is_liked) = &app.current_playback_is_liked {
+                if is_liked.to_owned() {
+                    "[liked]"
```
Rigellute: I was thinking we could show this character for "liked" songs
```
♥
```

svenvNL:
I would prefer to have all track ids that are liked (e.g. saved by the user), but we can only get 50 tracks per request. If we want to have all the tracks in a hashset, the app would have to make 27 request to get them all for my library. This check is only done once a new song starts playing, so I'm not sure it it worth the cost it has to booting the app.

Rigellute:
We would only make the request whenever we fetch new tracks. On first load, we can just check if the playing song is liked.

For example, when you open the "Liked Songs" view, we fetch the songs first and then make a request with all those ids to see if that page of songs contains any liked tracks.

We would do this with all the other views too.

Sorry if I'm not being clear. I can try to implement it this weekend.

svenvNL:
This would work for the first 20 songs, but when the Spotify plays song 21 (or later) and the user has not gone to page 2, there will be no liked / not liked data available. 

I have done the refactorings you asked for in https://github.com/svenvNL/spotify-tui/tree/playbar-liked-refactor. If my concern is not valid, I will see if I can finish it this weekend and merge it in the branch for this PR.

Rigellute:
I will try to implement my idea for saved tracks tomorrow. 

Not sure I've explained it so well but it's essentially this
- Fetch some tracks (e.g. when getting album tracks, playlist tracks, etc.) as per normal
- Get all those track ids and make spotify request to see if any are liked
- Spotify returns whether, for each track in the list, if it's liked or not
- We add the liked ones to the hashset.
- Then the UI track components just look up their id in this hashset and if it's there, show the heart .

Rigellute:
This is what I was trying to describe https://github.com/Rigellute/spotify-tui/pull/82

Rigellute:
Thanks again for taking a crack at this @svenvNL.

But will close this in favour of #82

PR #81, opened 2019-10-18
RFC: Use Vec<char> for input instead of String

fangyi-zhou:
This may make it easier to handle inputs since we do not have to care
about character boundaries in Strings.

See #30

I'm not sure whether this is a good idea performance-wise, I would like
to hear your opinions

### DIFF
--- a/src/handlers/input.rs
+++ b/src/handlers/input.rs
@@ -34,7 +34,7 @@ pub fn handler(key: Key, app: &mut App) {
             }
         }
         Key::Right => {
-            if app.input_cursor_position < app.input.len().try_into().unwrap() {
+             if app.input_idx < app.input.len() {
                 let next_c = app.input[app.input_idx];
                 app.input_idx += 1;
                 let width: u16 = UnicodeWidthChar::width(next_c).unwrap().try_into().unwrap();
--- a/src/handlers/input.rs
+++ b/src/handlers/input.rs
@@ -10,13 +10,14 @@ use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
 pub fn handler(key: Key, app: &mut App) {
     match key {
         Key::Ctrl('u') => {
-            app.input = String::new();
+            app.input = vec![];
             app.input_idx = 0;
             app.input_cursor_position = 0;
         }
         Key::Ctrl('e') => {
             app.input_idx = app.input.len();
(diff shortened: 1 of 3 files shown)


src/handlers/input.rs
```diff
@@ -26,15 +27,15 @@ pub fn handler(key: Key, app: &mut App) {
         }
         Key::Left => {
             if !app.input.is_empty() && app.input_idx > 0 {
-                let last_c = app.input.chars().nth(app.input_idx - 1).unwrap();
+                let last_c = app.input[app.input_idx - 1];
```
Rigellute: Could there ever be a situation where `app.input_idx - 1` goes out of bounds?

Perhaps it's safer to do `app.input.get(app.input_idx - 1)` and then handle the `Option`?

src/handlers/input.rs
```diff
@@ -26,15 +27,15 @@ pub fn handler(key: Key, app: &mut App) {
...
                 app.input_idx -= 1;
                 let width: u16 = UnicodeWidthChar::width(last_c).unwrap().try_into().unwrap();
                 app.input_cursor_position -= width;
             }
         }
         Key::Right => {
             if app.input_cursor_position < app.input.len().try_into().unwrap() {
-                let next_c = app.input.chars().nth(app.input_idx).unwrap();
+                let next_c = app.input[app.input_idx];
```
Rigellute: Same as above

src/handlers/input.rs
fangyi-zhou: it is guarded by the condition `app.index_idx > 0` so it's safe

src/handlers/input.rs
```diff
@@ -26,15 +27,15 @@ pub fn handler(key: Key, app: &mut App) {
...
             if !app.input.is_empty() && app.input_idx > 0 {
-                let last_c = app.input.chars().nth(app.input_idx - 1).unwrap();
+                let last_c = app.input[app.input_idx - 1];
                 app.input_idx -= 1;
                 let width: u16 = UnicodeWidthChar::width(last_c).unwrap().try_into().unwrap();
                 app.input_cursor_position -= width;
             }
         }
         Key::Right => {
             if app.input_cursor_position < app.input.len().try_into().unwrap() {
```
fangyi-zhou:
```suggestion
            if app.input_idx < app.input.len() {
```

Rigellute:
@all-contributors please add @fangyi-zhou for code

PR #97, opened 2019-10-22
Add info about libssl and pkf-config requirement

Bios-Marcel:
It's quite annoying that the compilation fails and you don't know about this upfront. Therefore I added generic information about the requirements.


Rigellute:
@all-contributors please add @Bios-Marcel for doc

Issue #69, opened 2019-10-15
Windows Installation


roketman09:
Hi,

About spotify-tui windows installation, there is a tutorial to use it? 

Greetings

Rigellute:
Hi @roketman09, I've not tried installing on Windows yet.

What issues are you encountering?

nathanloisel:
Windows installation doesn't work. After investigation, Termion dependency doesn't support Windows.
Here is the error i have:
```
error[E0433]: failed to resolve: maybe a missing `extern crate sys;`?
  --> C:\Users\Main\.cargo\registry\src\github.com-1ecc6299db9ec823\termion-1.5.3\src\lib.rs:24:9
   |
24 | pub use sys::size::terminal_size;
   |         ^^^ maybe a missing `extern crate sys;`?

error[E0433]: failed to resolve: maybe a missing `extern crate sys;`?
  --> C:\Users\Main\.cargo\registry\src\github.com-1ecc6299db9ec823\termion-1.5.3\src\lib.rs:25:9
   |
25 | pub use sys::tty::{is_tty, get_tty};
   |         ^^^ maybe a missing `extern crate sys;`?

error[E0433]: failed to resolve: maybe a missing `extern crate sys;`?
 --> C:\Users\Main\.cargo\registry\src\github.com-1ecc6299db9ec823\termion-1.5.3\src\async.rs:5:5
  |
5 | use sys::tty::get_tty;
  |     ^^^ maybe a missing `extern crate sys;`?

error[E0433]: failed to resolve: maybe a missing `extern crate sys;`?
  --> C:\Users\Main\.cargo\registry\src\github.com-1ecc6299db9ec823\termion-1.5.3\src\raw.rs:29:5
   |
29 | use sys::attr::{get_terminal_attr, raw_terminal_attr, set_terminal_attr};
   |     ^^^ maybe a missing `extern crate sys;`?

error[E0432]: unresolved import `sys`
  --> C:\Users\Main\.cargo\registry\src\github.com-1ecc6299db9ec823\termion-1.5.3\src\raw.rs:28:5
   |
28 | use sys::Termios;
   |     ^^^ maybe a missing `extern crate sys;`?

error[E0425]: cannot find function `get_tty` in this scope
  --> C:\Users\Main\.cargo\registry\src\github.com-1ecc6299db9ec823\termion-1.5.3\src\async.rs:14:36
   |
14 |     thread::spawn(move || for i in get_tty().unwrap().bytes() {
[...]

PR #99, opened 2019-10-22
Using cossterm for windows support.

TimonPost:
Fixes: https://github.com/Rigellute/spotify-tui/issues/69

This app can support windows very easily. TUI has the [crossterm backend](https://github.com/fdehau/tui-rs/blob/master/Cargo.toml#L33) which is made to support cross-platform. 

I tried to move this library over to crossterm just to experiment to how easy it was to swap termion. It was very easy. Please have a look at this [fork](https://github.com/TimonPost/spotify-tui/tree/crossterm_impl). 

Although... 

Their implementation uses crossterm 0.9 and 0.12 is already out and they didn't even release my latest patch - done a few months ago - with 0.10 yet. This old version of crossterm has some problems and ideally it should use 0.12 it's way better performance-wise as well.

However, I encountered problems:

-. `KeyEvent` (replacement for `Key`) does not implement `Copy` however, this is going to be fixed in 0.12.2. 
- TUI uses crossterm ^0.9, therefore, use crossterm ^12 because TUI requires types from ^0.9 
- You probably don't want to have a dependency to git branch of TUI

The real blocker here was that `crossterm::KeyEvent` did not implement `Copy` and therefore it has to be wrapped in order for it to be used.

Introduced:
1. `Key` enum, this enum is used in the library and should be used instead of the `termion::Key` or `crossterm::KeyEvent` or others toughtout the crates. You probably don't want to depend on a certain library restricted to linux but have a small abstraction so that crossterm, for example, can be implemented with more ease.
2. `CrosstermBackend` this uses crossterm for its implementation and is crossplatform. 

Qustions:

This is just an experiment because I wanted to listen to Spotify on windows, however, there are some concerns with git dependencies and old versions of crossterm. Do you think crossterm should be behind a feature flag, the only supported implementation, or not used at all?


TimonPost:
I tried this app out, however something went wrong with my authentication URL redirect. Not sure if anything went wrong with this implementation or authentication.

Rigellute:
Thanks for looking into this @TimonPost! This is great work, I'd love for `spotify-tui` to support Windows natively.

I think your `Key` abstraction is a good idea.

To respond to your questions above, 
1. I would certainly prefer to use crate releases and not use git as the dependancy manager here. 
1. If Crossterm delivers that same experience for all platforms as what we have already for macOS and Linux then it should be the only supported implementation, just to keep things simple. 

I'm now thinking that we should wait for/push `tui-rs` to upgrade their crossterm dependancy to `0.12`. And/or also wait for `crossterm` to support `Copy` (if that's still an issue).

But sounds like we're pretty close to supporting Windows, so once again thank you for this investigation!

TimonPost:
left an issue here: https://github.com/fdehau/tui-rs/issues/192

TimonPost:
I updated to the newest version of crossterm 0.14. We will have to wait for TUI to merge and release [this branch](https://github.com/fdehau/tui-rs/pull/209). Then this PR can finally be merged.

TimonPost:
image

TimonPost:
LINUX MINT
image

TimonPost:
* Remove /src/utils.rs,  moved logic to /events/*
* Created `Key` enum to be decoupled from external libraries.
* Implemented crossterm for UI driver.
* Upgraded to TUI 0.8 needed for crossterm 0.14 support.

TimonPost:
Debian:
image

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Rigellute: Are you able to remove the `.idea` directory?

TimonPost:
Nice, merging #190 sounds like a good idea. I made that one because this one accidentally included those changes as well. By merging that one, we can reduce the number of changes, which will make it a bit easier to see what the actual changes are.

.idea/vcs.xml
TimonPost: Certainly.

TimonPost:
I rebased and removed the idea files. It should be ready to be merged.

Rigellute:
@TimonPost have run into a problem after merging - pressing `ctrl-c` to quit, causes an error
```
Error: RecvError
```
Which then crashes that terminal session.

Any ideas?

TimonPost:
`RecvError` means that the send channel hung up.

TimonPost:
Something like: https://github.com/TimonPost/spotify-tui/commit/ca32947b1449d0426257c5d6470f44e1cf005c3c

PR #104, opened 2019-10-24
Unsave already saved songs with 's'

jfaltis:
Pressing s now toggles the saved state of a song, which allows unsaving songs. Kept the old save_tracks despite it not being used for usage in future changes.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -707,6 +707,33 @@ impl App {
         }
     }
 
+    pub fn toggle_save_track(&mut self, track_id: String) {
+        if let Some(spotify) = &self.spotify {
+            match spotify.current_user_saved_tracks_contains(&[track_id.clone()]) {
+                Ok(saved) => {
+                    if saved[0] {
+                        match spotify.current_user_saved_tracks_delete(&[track_id]) {
+                            Ok(()) => {}
+                            Err(e) => {
+                                self.handle_error(e);
+                            }
+                        }
+                    } else {
+                        match spotify.current_user_saved_tracks_add(&[track_id]) {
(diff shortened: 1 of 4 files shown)


alichtman:
Does it make sense to add a confirmation prompt for this potentially destructive action?

src/app.rs
```diff
@@ -707,6 +707,35 @@ impl App {
         }
     }
 
+    pub fn toggle_save_track(&mut self, track_id: String) {
+        if let Some(spotify) = &self.spotify {
+            match spotify.current_user_saved_tracks_contains(&[track_id.clone()]) {
+                Ok(saved) => {
+                    if saved[0] {
```
Rigellute: This is probably overkill, but to be safe let's use `saved.first()` instead of the array accessor `[0]`. We'd then need to handle the option.

src/app.rs
```diff
@@ -707,6 +707,35 @@ impl App {
...
+                }
+                Err(e) => {
+                    self.handle_error(e);
+                }
+            }
+        };
+    }
+
+    // Currently unused but keep for future changes
+    #[allow(dead_code)]
```
Rigellute: Are you able to delete this? No need to allow dead code (shouldn't exist in master).

Rigellute:
@alichtman that's an interesting point - let's see how it goes without a prompt. 

The official Spotify client doesn't throw up a prompt when removing saved tracks, so let's copy that behaviour for now.

jfaltis:
@Rigellute I adressed your changes. I also agree that we don't need to confirm this action since it can be made undone very easily

Issue #102, opened 2019-10-23
Application just closes when I am not using the PC


Bios-Marcel:
I usually leave the client open all day. As soon as i take my lunch break, I leave my PC and leave `spt` open. However, I do stop playback on my Spotify Connect device, which is my phone, which is the only active Spotify Connect device. When I come back, `spt` has exited without printing an error or anything.

Bios-Marcel:
Is there a logfile or anything?

jfaltis:
@Bios-Marcel I added the output of a backtrace recently when running spotify-tui in Debug mode (https://github.com/Rigellute/spotify-tui/pull/111). Could you follow these instructions (https://github.com/Rigellute/spotify-tui#development) and reproduce this crash?

Bios-Marcel:
Will do tmr if I find time

Bios-Marcel:
I am assuming this is related to sleep or standby? Couldn't such errors just be ignored and display something like "Connected severed. .. trying to reconnect."?

```
~/code/spotify-tui$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 4.29s
     Running `target/debug/spt`
Error Os { code: 2, kind: NotFound, message: "No such file or directory" };Please navigate here ["https://accounts.spotify.com/authorize?redirect_uri=http:%2F%2Flocalhost:8888%2Fcallback&state=bf7aMrWWTz9XOhrS&client_id=40717002b30f4887a92a804e5b7f2682&response_type=code&scope=playlist-read-collaborative%20playlist-read-private%20user-follow-read%20user-library-modify%20user-library-read%20user-modify-playback-state%20user-read-currently-playing%20user-read-playback-state%20user-read-private%20user-read-recently-played&"]
thread '<unnamed>' panicked at 'send request failed: Error(Io(Custom { kind: TimedOut, error: StringError("timed out") }), "https://accounts.spotify.com/api/token")', src/libcore/result.rs:999:5
stack backtrace:
   0: spt::panic_hook
             at src/main.rs:95
   1: spt::main::{{closure}}
             at src/main.rs:109
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::continue_panic_fmt
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
             at /usr/src/rustc-1.36.0/src/libcore/macros.rs:18
   7: core::result::Result<T,E>::expect
[...]

PR #111, opened 2019-10-25
Show backtrace on app crash

jfaltis:
This PR enables the output of the backtrace to the terminal when the app panics. This feature is currently only enabled in **Debug** and also ignores the `RUST_BACKTRACE` environment variable. It uses the workaround suggested here fdehau/tui-rs#177.

Unfortunately there is still an issue with the cursor placement after the crash where it is placed into the stacktrace output if the app was not run from the last line in terminal. Demonstration:

Cursor not at bottom before app is started:
backtrace

Cursor at bottom before app is started (expected behaviour):
backtrace-1
 

I couldn't figure out a solution yet but I think this could be fixed in a later PR because this is a valuable addition to track down the reason for crashes. This would improve situations like these https://github.com/Rigellute/spotify-tui/issues/102.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,10 +50,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "backtrace"
-version = "0.3.34"
+version = "0.3.40"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Rigellute:
@all-contributors please add @jfaltis for code

Issue #46, opened 2019-10-11
Implement a configuration file (e.g. for keybindings, etc.)


jfaltis:
Create a config for easy customization of keybindings, program behaviour, etc. It could be located at `$HOME/.config/spotify-tui/config.yml.`

The current client configurations could be moved into this configuration too.
E.g

```
---
client:
  - client_id:
  - client_secret:
  - device_id:
  
keybindings:
  - left: h
  - down: j
  - up: k
  - right: l
```

Rigellute:
This could be an interesting one.

Are there any bindings that don't currently work for you?

jfaltis:
I think the current bindings are very nice and appeal to the majority of users. But there are also users who might have other preferences (as you could see about the bindings for playback control https://github.com/Rigellute/spotify-tui/issues/16). Having a config for keybindings would be a nice addition for users who are into customizing everything as much as they can.

My suggestion would be that the default settings/bindings (which satisfy the most users) stay in the source code and the config which is being read can overwrite certain bindings. That way the config file doesn't contain all bindings but only the ones users want to change.

Currently I would not merge the client settings into the config yet and get rid of the client.yml. I think it would be nice if you could set these in the config.yml but the client.yml should stay to make it less of a hassle for users who sync their dotfiles with other computers. This is stuff for a later PR anyways.

I think I would rename `src/config.rs` to `src/client_config.rs` and add a `src/app_config.rs` for keybindings and preferences (etc.).

What do you think?

PR #112, opened 2019-10-25
Add keybinding config file

zacklukem:
This will add a `config.yml` file along side the `client.yml` file which
will allow the user to configure custom keybindings.  Currently this
commit only covers global keybindings and not per-screen keybindings but
this could be added in the future. This commit also doesn't currently
allow for overriding default navigation keys (hjkl etc.) but that could
also be added later.

Resolves: #46

### DIFF
--- a/src/user_config.rs
+++ b/src/user_config.rs
@@ -52,6 +52,31 @@ fn parse_key(key: String) -> Result<Key, failure::Error> {
     }
 }
 
+fn check_reserved_keys(key: Key) -> Result<(), failure::Error> {
+    let reserved = [
+        Key::Char('h'),
+        Key::Char('j'),
+        Key::Char('k'),
+        Key::Char('l'),
+        Key::Up,
+        Key::Down,
+        Key::Left,
+        Key::Right,
+        Key::Backspace,
+        Key::Char('\n'),
+    ];
+    for item in reserved.iter() {
+        if key == *item {
+            // TODO: Add pretty print for key
+            return Err(failure::format_err!(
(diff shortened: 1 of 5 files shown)


zacklukem:
I added code to check hjkl, return and arrow keys

src/user_config.rs
```diff
@@ -0,0 +1,236 @@
...
+                    "The key \"{}\" is unknown.",
+                    sections[0]
+                )),
+            }
+        }
+    }
+}
+
+fn check_reserved_keys(key: Key) -> Result<(), failure::Error> {
+    let reserved = [
```
zacklukem: We can add more keys to check here, but if we want to add checks for keys that the user can re-bind, we would need to use a vec or maybe a hashmap to be able to not block the user from doing something like this:
```yaml
decrease_volume: '+'
increase_volume: '-'
```
Because with current code, if we add `+` and `-` to the reserved keys, the User wouldn't be able to do something like this, even if it is valid.

I think it isn't too important to block keys that are configurable from being reused by the user at all, as those keys are documented in the README.

Thoughts?

jfaltis:
This looks very nice @zacklukem. I would suggest to move all the keybindings into one section in the config file to get a nicer structure into the config (because there are probably more settings sections than only keybindings in the future). This needs to be added asap so there is no need to transform existing configuration files into the new configuration file structure. I submitted a pull request with these changes to your branch.

Rigellute:
@all-contributors please add @zacklukem for code

Issue #11, opened 2019-10-07
Unable to scroll past first page of songs in playlist


VerbTheNoun95:
After selecting a playlist, I can only view the first 30 or so songs and can't scroll beyond that. Scrolling using vim keys or arrow keys brings me back to the top of the list, and ctl+d/u doesn't do anything (although I'm not too sure ctl+d/u should do). Hopefully I'm not missing something easy in the help section.

Rigellute:
You're right that `ctrl+d/u` should scroll through pages, but this has only been implemented in the "Liked Songs" view.

Will implement scrolling in other views!

daphen:
Or maybe you should be able to just scroll further down the list like in the "playlists-view"? Instead of jumping up when you get to the bottom I mean?

jessebett:
I agree with @daphen the behaviour inside the `Songs` view should be the same as the behaviour inside the `Playlists` view, not like the current `Liked Songs`. `Liked Songs` should also behave like this.

That is, once the `Songs` view has been selected,hitting `j` to the bottom should not cycle to the top, but should continue going down the view, like `Playlists`. 

`ctrl+d/u` will still be useful for scrolling up and down whole pages.

nikolaevigor:
Hi, what if I have `ctrl-d` bound to i3wm dmenu, can i remap it somehow? If yes, what key should I use in `keybindings:` part of `client.yml`?

Rigellute:
@nikolaevigor in the next release you'll be able to configure `ctrl-d/u` through the keybindings config https://github.com/Rigellute/spotify-tui/pull/566

PR #128, opened 2019-11-03
Add support for scrolling up/down when album list is active

echoSayonara:
This PR is related to #11

`ctrl+d/u` should scroll through pages but this has been implemented in "Liked Songs" view only. In this PR, I tried to implement scrolling down/up in the "Albums" view.

need to confirm:
In `src/app.rs`, I added three methods (`get_current_user_saved_albums`, `get_current_user_saved_albums_next`, and `get_current_user_saved_albums_previous`) for `App`. These methods are imitation of `**saved_tracks**` methods. Although `push_navigation_stack` is called in `get_current_user_saved_tracks`, I did **NOT** call `push_navigation_stack` in `get_current_user_saved_albums`.  If `push_navigation_stack` is called after each page scrolling, same contents are pushed into the stack. Scrolling pages several time, q key (go back) seems foul-up. This behavior seems a bug for me so I did not call `push_navigation_stack` in `get_current_user_saved_albums`.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -868,7 +868,7 @@ impl App {
             match spotify.current_user_saved_albums(self.large_search_limit, offset) {
                 Ok(saved_albums) => {
                     // not to show a blank page
-                    if saved_albums.items.len() > 0 {
+                    if !saved_albums.items.is_empty() {
                         self.library.saved_albums.add_pages(saved_albums);
                     }
                 }
--- a/src/app.rs
+++ b/src/app.rs
@@ -862,4 +862,43 @@ impl App {
             };
         };
     }
+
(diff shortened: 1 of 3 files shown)


Rigellute:
Thanks for taking a crack at this.

I think I've got things into a bit of a mess with regards to pagination (managing page index, scroll, index etc.).

There's a half baked attempt to clean this up, using a more generic struct (see `ScrollableResultPages` struct in `app.rs`).

I've got a better understanding of how this should work now and do intend to clean this up.

With regards to your comments, you're right about not pushing to the stack when fetching the next page. Pushing to the stack only needs to happen the first time we want to navigate to that view.

Issue #106, opened 2019-10-24
Wrong album is shown if an album in the Album library was viewed previously 


jfaltis:
Steps to reproduce
- launch `spotify-tui`
- select an album of your Albums in the Library
- select an artist of your Library and one of the artists albums

sputnick1124:
It seems that when an album is selected from an artist, It's track list is overwritten immediately by the track listing of the most recently viewed album from the user's saved albums list.

https://github.com/Rigellute/spotify-tui/blob/0379b46fab2c3a7ac85234f437e635c85998e4ba/src/handlers/artist_albums.rs#L29-L49

@Rigellute, is there a reason for this second block? Going through the commit history, it seems that the second block in that `artist_albums` handler was an earlier prototype which should perhaps have been removed in a later commit when the actual artist results were fleshed out?

Early commit:
https://github.com/Rigellute/spotify-tui/commit/e4f446535a838141e7ccf137f6d7b5559fa1810e#diff-f935c27baecb8324a6640180293f9d9eR27-R36
Later commit:
https://github.com/Rigellute/spotify-tui/commit/28e9ee28fd0933d95f8ed675f551e07cbd8b8a4e#diff-f935c27baecb8324a6640180293f9d9eR29-R39

sputnick1124:
I should mention that the lines causing the issue in the `artist_albums` handler are an exact copy of the lines performing the same function in the `album_list` handler, which is what leads me to believe that perhaps this is leftover prototyping code.

Problematic code:
https://github.com/Rigellute/spotify-tui/blob/0379b46fab2c3a7ac85234f437e635c85998e4ba/src/handlers/artist_albums.rs#L40-L49
Copied code:
https://github.com/Rigellute/spotify-tui/blob/0379b46fab2c3a7ac85234f437e635c85998e4ba/src/handlers/album_list.rs#L27-L36

sputnick1124:
I've got a [branch](https://github.com/sputnick1124/spotify-tui/tree/106-wrong-album) with the fix implemented. I can submit a PR unless I'm missing something.

Rigellute:
Great debugging @sputnick1124. You're right, this bug is a result of hacky prototyping. I really need to rethink and tidy things up.

I've gone through a few different approaches and it makes things confusing!

Please do PR your fix!

sputnick1124:
[...]

PR #130, opened 2019-11-04
fix: display correct artist album track listing

sputnick1124:
Addresses https://github.com/Rigellute/spotify-tui/issues/106


sputnick1124:
Whoops, not sure what to do about the failing test here... I still don't quite have the context of the full app to fully understand what is being tested or if it would be safe to remove the test. Do I need to update it? Bringing in `ActiveBlock` just for the test would of course help the test along, but seems not to be the most sensible approach?

Any advice here would be appreciated. Thanks!

PR #133, opened 2019-11-04
Add support for page scrolling in Playlist

### LINKED ISSUE #11: Unable to scroll past first page of songs in playlist

ghost:
Fix #11 

I tried to implement page scrolling in the Playlist and fixed `get_current_user_saved_tracks` that I pointed out in #128 .

I added and changed filed and methods of `App` but I'm not sure whether these changes are matched to what you expected. I would like to hear your opinion.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -685,7 +685,6 @@ impl App {
 
                     self.library.saved_tracks.add_pages(saved_tracks);
                     self.track_table.context = Some(TrackTableContext::SavedTracks);
-                    self.push_navigation_stack(RouteId::TrackTable, ActiveBlock::TrackTable);
                 }
                 Err(e) => {
                     self.handle_error(e);
--- a/src/app.rs
+++ b/src/app.rs
@@ -219,7 +219,8 @@ pub struct App {
     pub large_search_limit: u32,
     pub library: Library,
     pub playback_params: PlaybackParams,
-    pub playlist_tracks: Vec<PlaylistTrack>,
+    pub playlist_tracks: ScrollableResultPages<Page<PlaylistTrack>>,
+    pub playlist_id: Option<String>,
     pub playlists: Option<Page<SimplifiedPlaylist>>,
     pub recently_played: SpotifyResultAndSelectedIndex<Option<CursorBasedPage<PlayHistory>>>,
     pub search_results: SearchResult,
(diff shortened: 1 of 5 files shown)


src/app.rs
```diff
@@ -219,7 +219,8 @@ pub struct App {
     pub large_search_limit: u32,
     pub library: Library,
     pub playback_params: PlaybackParams,
-    pub playlist_tracks: Vec<PlaylistTrack>,
+    pub playlist_tracks: ScrollableResultPages<Page<PlaylistTrack>>,
```
Rigellute: This is what I was thinking 👍

Rigellute (approved):
Great job again @echoSayonara.

This looks good to me 👍 

Perhaps once pagination has been implemented for all other views, a cleaner abstraction might be revealed.

Rigellute:
@echoSayonara I've reverted the merge as playback doesn't seem to quite work when trying to play tracks from other pages.

Do you want to have a look?

Rigellute:
I have a quick look and came up with this https://github.com/Rigellute/spotify-tui/compare/revert-133-page-scroll-rigel?expand=1

One issue here is that after the selection index gets above ~23, there is an "off-by-one" problem i.e. you press enter on the track, but the next one starts playing.

Could this be a problem with `usize` arithmetic?

echoSayonara:
Looks good.
I checked "off-by-one" error and it did not occur.

However, I found another problem.
After open one playlist, I open another playlist and press enter on the track, but a track in other page was played. It seems page index is shared with all playlists.

bdavs3:
**System**: MacOS | **Terminal**: KittyTerm + Zsh

@Rigellute I just installed `spotify-tui` for the first time. What I'm seeing with page scroll behavior in playlists is that pressing ctrl-d skips a page when it shouldn't. For instance:

* If I press `ctrl-d` on a playlist with 15 tracks, I can never access tracks 11-15
* If I press `ctrl-d` on a playlist with 45 tracks, I will skip from tracks 1-10 to tracks 21-30

Do you see this behavior as well? Want me to try to fix it?

bdavs3:
Looks like this is being talked about in #362 actually

Issue #10, opened 2019-10-07
scroll not in view


estysdesu:
When moving cursor up and down, it doesn't show selected/highlighted choice when terminal window is resized. I use tmux, and the pane resizes great to fit the proper space, but it doesn't update the scroll buffer of what is visible to select. 
I am selecting songs below the visible song list, because they are still selectable in the current scroll buffer from the tmux resize.

![Scroll not in view](https://user-images.githubusercontent.com/37026871/66340711-c927f400-e913-11e9-8171-b1ba3ed11a75.gif)

Rigellute:
Ah good point, this needs to be fixed.

sputnick1124:
I can take this if no one else is working on it!

Rigellute:
Please do @sputnick1124 
I think the fix would entail figuring out how many rows would fit in the current window and also using that number as the spotify limit for results.

Rigellute:
Fixed by @sputnick1124

PR #135, opened 2019-11-04
10 scroll to selected item

sputnick1124:
First crack at addressing https://github.com/Rigellute/spotify-tui/issues/10.

Since tui-rs' `Table` widget does not automatically scroll to the seleted item automatically, I took some inspiration from the [`SelectableList` widget](https://github.com/fdehau/tui-rs/blob/master/src/widgets/list.rs#L213-L239) which does.

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -1097,7 +1097,16 @@ fn draw_table<B>(
 
     let (title, header) = table_layout;
 
-    let rows = items.iter().enumerate().map(|(i, item)| {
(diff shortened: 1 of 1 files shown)


sputnick1124:
Thanks! I love finding all of the very useful behaviors attached to iterators. That and sum types are two of the main reasons I love rust so much!

Good idea about requesting scrolling tables upstream! I'll look into filing an issue or even PR to `tui-rs` with the feature request.

PR #137, opened 2019-11-05
Fix paging in the Playlist

echoSayonara:
Fix the problem that a wrong track is played after open some playlists. https://github.com/Rigellute/spotify-tui/pull/133#issuecomment-549640154

Reset `playlist_tracks` when go back from the track table of a playlist and it fix the problem.

Paging in the playlist does not have problems that you pointed out and I found. Scrolling down/up works well and a correct track is played when i press enter in the track table of a playlist.

I commented out some codes that does not go well (line 249-256 in `main.rs`). Although it goes not work but it seems better to me. I do not know why those codes does work. Please check it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,7 +26,7 @@ use termion::screen::AlternateScreen;
 use tui::backend::{Backend, TermionBackend};
 use tui::Terminal;
 
-use app::{ActiveBlock, App};
+use app::{ActiveBlock, App, RouteId, ScrollableResultPages /* TrackTableContext */};
 use banner::BANNER;
 use config::{ClientConfig, LOCALHOST};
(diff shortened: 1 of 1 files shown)


Rigellute:
Hmm... this seems to add some complexity. I think we need to do a major overhaul of the state.

It's getting harder to know which bits of state are UI/data specific.

echoSayonara:
i agree. an overhaul is repuired.

indexes of tracks, albums, and pages are shared with some context and it causes errors, which unexpeted track is played.

it is not easy to solve these problems but it makes the codes clearer and this app useful.

i close this pr for the present and i open an issue about these problems.

Rigellute:
Thanks @echoSayonara. 

I need to carve out some time to dedicate to this problem

Issue #125, opened 2019-11-01
Panic when switching to short track from long current progress in longer track


sputnick1124:
# Reproduce
After playback on a long track has attained `x` seconds, switch to track with duration `y` where `y < x`

# Symptom
Panic
```
thread '<unnamed>' panicked at 'Percentage should be between 0 and 100 inclusively.', /home/nick/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.6.2/src/widgets/gauge.rs:47:9
stack backtrace:
...
   3: std::panicking::begin_panic
             at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c/src/libstd/panicking.rs:408
   4: tui::widgets::gauge::Gauge::percent
             at /home/nick/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.6.2/src/widgets/gauge.rs:47
   5: spt::ui::draw_playbar
             at src/ui/mod.rs:677
   6: spt::ui::draw_main_layout
             at src/ui/mod.rs:161z
...
```

The issue seems to be that the track duration and progress numbers are updated at different times/places, so the app tries to update the progress bar with previous track's duration over the new track's duration, resulting in a panic from tui-rs's `Gauge` widget.

Simply placing a bandage on the problem by limiting the percentage to 100% maximum elicits a panic in a different place for the same reason
```rust
            let perc = if app.song_progress_ms < track_item.duration_ms.into() {
                (app.song_progress_ms as f64 / f64::from(track_item.duration_ms)) * 100_f64
            } else {
                100.0
            };
```
```
thread '<unnamed>' panicked at 'attempt to subtract with overflow', src/ui/util.rs:52:39
stack backtrace:
...
   6: core::panicking::panic
             at src/libcore/panicking.rs:49
   7: spt::ui::util::display_track_progress
             at src/ui/util.rs:52
   8: spt::ui::draw_playbar
             at src/ui/mod.rs:690
...
```
 where https://github.com/Rigellute/spotify-tui/blob/06d7b91164f0640dbdf770113f623ec8aea5f4f2/src/ui/util.rs#L52
panics because of a negative unsigned result.

PR #139, opened 2019-11-06
fix: Update current progress at new track load

sputnick1124:
There was a race condition between selecting a new track and the track
progress updating; this caused the app to sometimes panic when
selecting a track which has a total duration less than the current
progress as it tries to report a negative remaining time in the playbar
(but it tries this with unsigned values).

```bash
/git/spot.. master λ › cargo run
   Compiling spotify-tui v0.8.0 (/home/user/git/spotify-tui)
    Finished dev [unoptimized + debuginfo] target(s) in 6.75s
     Running `target/debug/spt`
thread '<unnamed>' panicked at 'attempt to subtract with overflow', src/ui/util.rs:52:39
stack backtrace:
...
   6: core::panicking::panic
             at src/libcore/panicking.rs:49
   7: spt::ui::util::display_track_progress
             at src/ui/util.rs:52
   8: spt::ui::draw_playbar
             at src/ui/mod.rs:685
   9: spt::ui::draw_main_layout
             at src/ui/mod.rs:161
  10: spt::main::{{closure}}
             at src/main.rs:192
...
```

This, I think, was ultimately the root cause of https://github.com/Rigellute/spotify-tui/issues/125 and the solution we've been looking for with https://github.com/Rigellute/spotify-tui/pull/132 and https://github.com/Rigellute/spotify-tui/pull/126.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -563,6 +563,7 @@ impl App {
         match result {
             Ok(()) => {
                 self.get_current_playback();
+                self.song_progress_ms = 0;
                 self.playback_params = PlaybackParams {
                     context_uri,
                     uris,


Rigellute:
After learning about the `checked` arithmatic methods (e.g. `checked_sub`), perhaps we should use those here too?

sputnick1124:
I think that would be a good idea, just to guard against anything unforeseen! :sweat_smile: 

I also think that once we have implemented logging (if we go that route), that anything any of these checked operations fails, we definitely ought to log it so as not to lose insight into a perhaps more fundamental issue somewhere else in the app.

Rigellute:
@all-contributors please add @sputnick1124 for code

PR #141, opened 2019-11-06
Add Ctrl+g and G vim style keybindings to move cursor to first and last line

echoSayonara:
I add new keybindings for cursor motion.

ctrl+g and G are Vim-like keybinding that move cursor to first and last line of the list. Real vim keybinding to move cursor to first line is gg but it seems that  keybindings that use two or more key are not supported in this app so I assign the fiunction to ctrl+g.

I think that i implemented them in all blocks (Artists, Albums, Playlists and so on) but if you notice my over looking, please tell me it.

### DIFF
--- a/src/handlers/common_key_events.rs
+++ b/src/handlers/common_key_events.rs
@@ -1,4 +1,4 @@
-use super::super::app::{ActiveBlock, App, RouteId};
+use super::super::app::{ActiveBlock, AlbumTableContext, App, RouteId, SearchResultBlock};
 use termion::event::Key;
 
 pub fn down_event(key: Key) -> bool {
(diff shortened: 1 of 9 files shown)


src/handlers/common_key_events.rs
```diff
@@ -133,6 +133,141 @@ pub fn handle_left_event(app: &mut App) {
...
+            SearchResultBlock::Empty => {}
+        },
+        ActiveBlock::MyPlaylists => {
+            app.selected_playlist_index = Some(0);
+        }
+        _ => {}
+    }
+}
+
+pub fn go_last_line(app: &mut App) {
```
Rigellute: Eventually, we could probably improve things by having a more consistent way of managing the selection index. I'll look into this as part of working on better pagination.

Issue #138, opened 2019-11-05
tui not using space efficiently


iMarv:
The tui is not extending over the complete terminal window, leaving a lot of space empty.

![spotify_tui_screnshot](https://user-images.githubusercontent.com/2131339/68188503-95d79400-ffa9-11e9-954e-b2111d6fda05.png)

Rigellute:
That space at the bottom is meant for the playbar. When nothing is playing, the playbar doesn't render. 
I keep meaning to address this.

iMarv:
Even with the playbar there is a lot of unused space:

![spotify_tui_screnshot2](https://user-images.githubusercontent.com/2131339/68206664-997d1200-ffcd-11e9-8df1-ef1cfbe4b4b1.png)

synth-ruiner:
I second this. I don't tend to have terminals any bigger than 1/4 of the screen (tiling window manager...) and spotify-tui doesn't make the most of the space available. I can only fit 11 tracks on screen that way. having 2 whole rows of padding at the top and bottom seems excessive!

Rigellute:
As a quick fix, I could remove the margin entirely?

<img width="1017" alt="Screenshot 2019-11-06 at 09 30 24" src="https://user-images.githubusercontent.com/12150276/68285917-149b0280-0078-11ea-9183-1280c44c7829.png">

synth-ruiner:
looks awesome to me! I can see how fullscreen you might want a small margin but that could be a job for later? :)

PR #142, opened 2019-11-06
Added config module 'visuals' and added window margin configuration

jfaltis:
This PR references https://github.com/Rigellute/spotify-tui/issues/138. I added the functionality to configure through window margin in the configuration file.

E.g. to remove the margin add to `~/spotify-tui/config.yaml`

```
visuals:
    margin: 0
```

If the margin is unset its default value is 1 which I think looks better than the previous default of 2 if you want to keep the margin. @Rigellute what default margin do you prefer?

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,9 @@ keybindings:
     shuffle: 's'
     repeat: 'r'
     search: '/'
+
+visuals:
+    margin: 1
 ```
 
 ## Limitations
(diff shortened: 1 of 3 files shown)


Rigellute:
Thanks for look at this @jfaltis , however I was thinking of fixing this issue along these lines https://github.com/Rigellute/spotify-tui/pull/143

PR #148, opened 2019-11-13
Bump versions in Cargo.lock.

tem1029:
The app would hang upon trying to authenticate with Spotify
on FreeBSD, bumping the versions to the latest seems to have
fixed this issue. All tests pass and works as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adler32"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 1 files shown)


tem1029:
Any update on this? I've simply ran ``cargo update`` and that fixed the issue on FreeBSD, where during the process of opening the browsing then authenticating it would hang, of which seemed to be fixed by updating the packages deps, subsequently fixing that issue.

tem1029:
Awesome, thanks. When you've pushed a new tag I'll be resubmitting a FreeBSD port that I sent in a while back (but unfortunately at the time wouldn't work to the now fixed issue). :+1:

Rigellute:
If you know how we can add a package manager for FreeBSD (or any other linux distros), please let me know!

tem1029:
There seems to be this in the Arch Linux community repository ( https://aur.archlinux.org/packages/spotify-tui/ ), although I'm not sure the status of other Linux distros. Might be worth nothing this in the README or something similar as an alternative installation method.

PR #150, opened 2019-11-14
Added pagnation to user playlists

Peterkmoss:
Added some simple pagnation to at least scroll through your playlist. I don't know if this is the best way to implement it but it seems to work fine from my own testing. Please excuse if my Rust is "rusty", it's my first Rust project and I don't know if there is need for memory optimization for this change to not leak memory.

### DIFF
--- a/src/handlers/track_table.rs
+++ b/src/handlers/track_table.rs
@@ -112,7 +112,7 @@ pub fn handler(key: Key, app: &mut App) {
                             if let Some(selected_playlist) =
                                 playlists.items.get(selected_playlist_index.to_owned())
                             {
-                                app.playlist_offset += 20;
+                                app.playlist_offset += app.large_search_limit;
                                 let playlist_id = selected_playlist.id.to_owned();
                                 app.get_playlist_tracks(playlist_id);
                             }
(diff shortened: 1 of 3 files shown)


src/handlers/track_table.rs
```diff
@@ -105,7 +105,19 @@ pub fn handler(key: Key, app: &mut App) {
...
                 Some(context) => match context {
-                    TrackTableContext::MyPlaylists => {}
+                    TrackTableContext::MyPlaylists => {
+                        if let (Some(playlists), Some(selected_playlist_index)) =
+                            (&app.playlists, &app.selected_playlist_index)
+                        {
+                            if let Some(selected_playlist) =
+                                playlists.items.get(selected_playlist_index.to_owned())
+                            {
+                                app.playlist_offset += 20;
```
Rigellute: Instead of hardcoding `20`, please use `self.large_search_limit`

Peterkmoss: I see, of course - I have fixed it to use the `large_search_limit` from app.

Peterkmoss:
I spotted a problem with the navigation when paging so I added a check to see if the new table should be added to the navigation stack.

Rigellute:
Thanks for again for the contribution

Rigellute:
@all-contributors please add @Peterkmoss for code

PR #151, opened 2019-11-17
Go back does not work in Liked Songs

echoSayonara:
Go back `q` does not work as expected after scrolling pages in Liked Songs.
When i scrolled pages in Liked Songs and pushed `q`, i expected that the Library block is activated but nothing occurred.

This problem caused by calling push_navigation_stack in get_current_user_saved_tracks.
I tried to fix this problem by removing push_navigation_stack from get_current_user_saved_tracks and pushing to the stack is called first time.


echoSayonara:
As you said, i merged branches to master and they contaminated it. I cleaned up unnecessary commits.

PR #152, opened 2019-11-17
Delete saved album

echoSayonara:
In the roadmap, delete saved contents is essential but it is not implemented yet.
Here, I tried to implement delete saved album and it seems to work well.

I assigned `D` to delete an album in album list but i do not know this key mapping is good for you and if you do not like this, please change key mapping.

### DIFF
--- a/src/handlers/album_list.rs
+++ b/src/handlers/album_list.rs
@@ -37,6 +37,7 @@ pub fn handler(key: Key, app: &mut App) {
         }
         Key::Ctrl('d') => app.get_current_user_saved_albums_next(),
         Key::Ctrl('u') => app.get_current_user_saved_albums_previous(),
+        Key::Char('D') => app.delete_current_user_saved_album(),
         _ => {}
     };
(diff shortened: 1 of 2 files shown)


src/handlers/album_list.rs
```diff
@@ -37,6 +37,7 @@ pub fn handler(key: Key, app: &mut App) {
         }
         Key::Ctrl('d') => app.get_current_user_saved_albums_next(),
         Key::Ctrl('u') => app.get_current_user_saved_albums_previous(),
+        Key::Char('D') => app.delete_current_user_saved_album(),
```
Rigellute: It is hard thinking about what the key mapping should be, but given that this is context specific to the album_list, I think this is good.

Rigellute:
Looks like there are some odd commits in this PR too, are you able to clean them up?
I think this should be one commit.

echoSayonara:
Cleaned up unnecessary commits.

PR #155, opened 2019-11-18
Follow/unfollow artists

echoSayonara:
Add support for following/unfollowing artists.

Considering #152, it seems that adding/following and deleting/unfollowing have common components and there may be more good abstraction. I will try this after adding and deleting were implemented in other type contents.

### DIFF
--- a/src/ui/help.rs
+++ b/src/ui/help.rs
@@ -76,5 +76,6 @@ pub fn get_help_docs() -> Vec<Vec<&'static str>> {
             "Pagination",
         ],
         vec!["Delete saved album", "D", "Library -> Albums"],
+        vec!["Follow an artists", "w", "Search result"],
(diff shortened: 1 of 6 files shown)


echoSayonara:
Added new bindings but some features are not implemented yet. I will soon implement them.

src/ui/help.rs
```diff
@@ -75,5 +75,15 @@ pub fn get_help_docs() -> Vec<Vec<&'static str>> {
             "<Ctrl+u>",
             "Pagination",
         ],
+        vec![
+            "Add selected artist, album, or track to the library",
```
Rigellute: For now, let's say only "Follow an artist"

Rigellute:
Great work @echoSayonara!

PR #156, opened 2019-11-19
add hotkey to copy url of currently playing track

LennyPenny:
Hey, 
first of all: Awesome project! 

I implemented a new feature that copies the url of the currently playing track when you press a button (`c` by default). 

This is my first time using rust, so if you find any mistakes or weirdness, let me know and I will fix the issues. 

I hope I found all the relevant sections I needed to adjust.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,6 +1,6 @@
 use super::config::ClientConfig;
 use super::user_config::UserConfig;
-use failure::err_msg;
+use failure::{err_msg, format_err};
 use rspotify::spotify::client::Spotify;
 use rspotify::spotify::model::album::{FullAlbum, SavedAlbum, SimplifiedAlbum};
 use rspotify::spotify::model::artist::FullArtist;
@@ -20,6 +20,9 @@ use rspotify::spotify::senum::{Country, RepeatState};
 use std::collections::HashSet;
 use std::time::Instant;
 use tui::layout::Rect;
+
+use clipboard::ClipboardContext;
+use clipboard::ClipboardProvider;
 
 pub const LIBRARY_OPTIONS: [&str; 6] = [
     "Made For You",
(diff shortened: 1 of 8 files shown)


LennyPenny:
it seems like the open/libre ssl depency is missing from the github action runner - any tips on how to add that to the github action runner?

src/main.rs
```diff
@@ -159,6 +160,11 @@ fn main() -> Result<(), failure::Error> {
 
             app.spotify = Some(spotify);
 
+            app.clipboard_context =
```
Rigellute: Could/should the `clipboard_context` be initialised in `app.rs`? Or is it easier to handle the error from `main`?

src/ui/help.rs
```diff
@@ -13,6 +13,7 @@ pub fn get_help_docs() -> Vec<Vec<&'static str>> {
         vec!["Seek backwards 5 seconds", "<", "General"],
         vec!["Seek forwards 5 seconds", ">", "General"],
         vec!["Toggle shuffle", "<Ctrl+s>", "General"],
+        vec!["Copy url to currently playing song", "c", "General"],
```
Rigellute: Nice one

src/app.rs
```diff
@@ -648,6 +653,24 @@ impl App {
         }
     }
 
+    pub fn copy_song_url(&mut self) {
+        let clipboard = match &mut self.clipboard_context {
```
Rigellute: Looks like this could be using an `if let`?

```rust
if let Some(clipboard_context) = &mut self.clipboard_context {
  ...
}
```

src/main.rs
LennyPenny: Problem is that currently app initialization cannot fail. If we make it failable we will have to adjust every test. That would touch almost every file, so I refrained from doing so

src/app.rs
LennyPenny: I used an "early return" here to save an indentation level, do you want me to change it?:)

LennyPenny:
regarding the ci runner: I think the `xorg-dev` package is missing from the runner

Rigellute:
Going to merge this into a side branch so I can test CI/CD fix you suggested

PR #162, opened 2019-11-24
Fix columns width in Saved Albums

gpawlik:
### Description of the issue
Massive PR! 😄 

The `Release Date` column was missing in the `Albums` table, which was caused by width exceeding the total available space.

The same happens for `Length` in `Recently Played Tracks` table, because of `Liked` column of fixed width `2`. I couldn't decide whether change it to a percentage calculated value or just subtract it from the next column (eg. `width: get_percentage_width(layout_chunk.width, 2.0 / 5.0) - 2`). What do you think @Rigellute ? 

| Screenshot 2019-11-24 at 21 16 03 | Screenshot 2019-11-24 at 21 15 32 |

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -943,7 +943,7 @@ where
             },
             TableHeaderItem {
                 text: "Release Date",
-                width: get_percentage_width(layout_chunk.width, 1.0 / 3.0),
+                width: get_percentage_width(layout_chunk.width, 1.0 / 5.0),
                 ..Default::default()
             },
         ],


gpawlik:
Cool @Rigellute I will fix the other issue in a separate PR. It is a pleasure to deal with this codebase! 😄

tem1029:
Hello! Since this commit I've noticed when you select a playlist, the header for Album has the text of "AlbumTrack", is this intended behavior? This can be seen here: **https://github.com/Rigellute/spotify-tui/blob/master/src/ui/mod.rs#L553**

Rigellute:
Good catch @tem1029. This is indeed a mistake, it should just say "Album".

I would guess this is a result of a find and replace 😄 .

Feel free to open a PR fixing it!

tem1029:
@Rigellute done so in here #165. Cheers.

Issue #101, opened 2019-10-23
Volume controll does not seem to work (correctly)


johandegn:
_Using version 0.7.0 and playing from spotify app on android._
Pressing '+' has no effect and '-' opens the error screen.
(Pressing '-' multiple times stacks error screens so you need to press Esc multiple times to exit.)

johandegn:
While it works then streaming from the desktop application on a windows 10 pc, the volume acts weird.
Switching between + and - makes it jump to a, seemingly, random percentage.
It also seems that the volume first changes to the shown after the next change. 
For example, say the volume is 65%. Pressing either + or - might make it jump to 37% but actually changing the volume to the 65%. Next change would make it jump to another number, but setting the volume to the 37%.

Rigellute:
The volume controls are quite janky. Will investigate.

Rigellute:
Fixed in https://github.com/Rigellute/spotify-tui/pull/166

PR #166, opened 2019-11-25
Adjust volume controls behaviour

gpawlik:
### Description of the issue
The volume control does not work as intended, if `+` and `-` are pressed in a moderate pace the value updates every second press, and after some time it starts to get even more messy.

### Solution to the issue
I have removed `get_current_playback` call after the successful volume change. In this context I think it is not necessary to poll for the playback data, but maybe I am missing something, would you be able to confirm it @Rigellute ?

### Before/After screenshots 📷

| bbb | aaa |


Rigellute:
This PR will close https://github.com/Rigellute/spotify-tui/issues/101

PR #167, opened 2019-11-25
Playlist: add shortcuts to jump to the start or the end

BlackYoup:
Hello,

Here's a patch to allow a user to jump to the end / start of a playlist.

I used the same shortcut as the one used to go at the end / start of the search input but I'm not sure if you are ok with that. If you are, I guess we'd have to find how to specify it in the help page.

I also didn't implement it for other variants of `TrackTableContext` as I'm not sure if that's worth it.

Let me know how you feel about this and thanks for your work on this client :)

### DIFF
--- a/src/handlers/track_table.rs
+++ b/src/handlers/track_table.rs
@@ -155,6 +155,66 @@ pub fn handler(key: Key, app: &mut App) {
                 None => {}
             };
         }
+        Key::Ctrl('e') => jump_to_end(app),
+        Key::Ctrl('a') => jump_to_start(app),
         _ => {}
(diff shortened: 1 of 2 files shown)


BlackYoup:
> I think your proposed bindings work well. Could you add these to help.rs please?

Alright, will do :)

> Seems like the linter "Clippy" has some issues. Will merge once these CI problems are solved.

Yes, I was waiting for your input to fix those in case you weren't ok with the changes, I'll fix it later today!

Rigellute:
@all-contributors please add @BlackYoup for code

PR #168, opened 2019-11-26
make seeking amount configurable

LennyPenny:
This adds the config value `seek_seconds` to change much `spotify-tui` seeks when pressing the seek hotkeys. 
Default is still `5000` ms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,8 @@ keybindings:
   shuffle: "s"
   repeat: "r"
   search: "/"
+
+seek_seconds: 5
 ```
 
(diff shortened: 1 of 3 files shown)


jfaltis:
This looks like a nice addition! I would suggest moving the seek_seconds into a "behavior" section to properly structure the config file (as already done with the keybindings). What do you think?

LennyPenny:
Maybe call it `general`?

Rigellute:
@all-contributors please add @LennyPenny for code

PR #170, opened 2019-11-26
Adjust min/max volume values

gpawlik:
### Description of the issue
Currently if the initial volume is not of full tens (eg. `42`) then the min and max values are respectively `2` and `92`.

### Solution to the issue
Limit the volume values correctly on `0` and `100`.

### Before/After screenshots 📷

| v2 | v1 |


gpawlik:
It's a pleasure! 😄

Rigellute:
@all-contributors please add @gpawlik for code

PR #177, opened 2019-12-01
Fix playlist index after search

echoSayonara:
After searching, playlist index is changed to `None` and it makes playlist selection odd.
I deleted a line that binds `None` to `selected_playlist_index` and it worked well.


echoSayonara:
You are right. Fixed it.

Issue #174, opened 2019-11-30
Typing 'q' to exit focuses search bar


stevensonmt:
After playing some tracks, typing q to get back to the main screen and then q to quit the application puts the cursor in the search bar. Hitting 'esc' and then 'q' typically then exits the application.

Rigellute:
Hi @stevensonmt, you are right. I had originally intended this to be that way, however I myself don't like it.

If anyone want to have a crack at fixing, we could possibly do this: when we are `pop`ing the navigation stack via `q` we can check if the `pop`ed route is `Search` and then immediately `pop` again - essentially bypassing the search route.

stevensonmt:
I can have a go at that this evening. Seems pretty straightforward to match on the pop result of Some(Search) and pop again.

samcal:
@Rigellute I think this issue can be closed as fixed.

Rigellute:
Thanks @samcal 👍

PR #184, opened 2019-12-02
Remove focus on input when jumping back

TheWalkingLeek:
This is fixing #174 

First of all I wanted to say that I'm pretty new to Rust so if you notice anything and have tips for me please feel free to share them with me :)

Also I realized that if the first action is to use the input field (thus the second route in the `navigation_stack` being a route with the id `Search`) the `Home` route will not be shown on navigating back because the second route will be popped in the match I've added and the `is_none()` check will break after...

I didn't know how to prevent this effectively without slaughtering the code. So any ideas would be really appreciated.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,7 +27,7 @@ use termion::screen::AlternateScreen;
 use tui::backend::{Backend, TermionBackend};
 use tui::Terminal;
 
-use app::{ActiveBlock, App};
+use app::{ActiveBlock, App, RouteId};
 use banner::BANNER;
 use config::{ClientConfig, LOCALHOST};
 use redirect_uri::redirect_uri_web_server;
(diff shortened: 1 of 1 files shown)


stevensonmt:
How about this:
```
let mut pop_result = app.pop_navigation_stack();
    match &pop_result {
        Some(route) => {
            pop_result = match route.id {
                RouteId::Search => app.pop_navigation_stack(),
                _ => pop_result
            }
        },
        None => {}
    }
if pop_result.is_none() {
    break; // Exit application
}
```
becomes this:
```
let pop_result = { match app.pop_navigation_stack() {
    Some(RouteId::Search) => app.pop_navigation_stack(),
    Some(x) => Some(x),
    None => None
};

if pop_result.is_none() {
    break // Exit application
}
```

TheWalkingLeek:
Yeah i see, makes it definitely more compact and readable. Will change it as soon as i get to it

TheWalkingLeek:
So I've tried your solution but it didn't quite add up. Just because on the first `pop_navigation_stack` call it returns an `Option<Route>` which we can't match on `Some(RouteId::Search)`. And unwrapping the Option would result in a panic on the last occurence.
I did like the idea of just one assignement though so I've still refactored it a bit. Let me know what you think

stevensonmt:
My apologies for not reading the original more closely. You could have done something like this to fix it:
```
let pop_result = { match app.pop_navigation_stack() {
    Some(x) if x.id == RouteId::Search  => app.pop_navigation_stack(),
    Some(x) => Some(x),
    None => None
};

if pop_result.is_none() {
    break // Exit application
}
```

TheWalkingLeek:
No need to apologize. Thanks for the help, didn't know about the guard clause inside a match arm. Tell me if somethings still not right. I've also just rebased and compressed to one commit

Rigellute:
@all-contributors please add @TheWalkingLeek for code

PR #188, opened 2019-12-09
Pad strings in status bar to prevent reformatting

jacobchrismarsh:
Currently there are different-length strings for parameters, such as
`On` and `Off`. When you switch options, the bottom status bar
reformats. This is a worse user experience, so this diff pads the bottom
status line so that it never reformats.

The included gifs show the end result.

### Old
old_spt

### New
improved_spt

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -638,7 +638,7 @@ where
             };
 
             let title = format!(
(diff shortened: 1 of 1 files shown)


LennyPenny:
Would it be possible to still make the items centered? It seems odd that they are not centered when the shorter `Off` is displayed instead of `Track`

Rigellute:
@all-contributors please add @jacobchrismarsh for code

PR #190, opened 2019-12-13
Merge imports

TimonPost:
Ran rustfmt merge imports. 

I really like this in my libraries. To me, this looks way cleaner.


TimonPost:
Not sure if that is possible. Probably you won't wan this tough, because it is a nightly feature. Adding it to the rustfmt.toml will cause warnings on stable.

Rigellute:
Okay cool 👍 .

Thanks @TimonPost

PR #199, opened 2019-12-25
Windows compatibility with crossterm

TimonPost:
Continues on #99 which had had an issue after merging. This PR contains the fix.

- Remove termion dependency
- Use crossterm feature flag TUI
- Added event module, for event handling behavior.
- Implemented crossterm 0.14.


Rigellute:
@all-contributors please add @TimonPost for code

Issue #192, opened 2019-12-16
Disallow scrolling past the end of a playlist


alichtman:
If you hammer on `Ctrl-D` for a while, you'll see that you can scroll past the end of a playlist. You need to press `Ctrl-U` as many times as you pressed `Ctrl-D` after the playlist had already been scrolled past in order to get back to the bottom.

I'd expect scrolling down to bottom out at the end of a playlist.

Rigellute:
You're right. Needs fixing 👍

Monkeyanator:
Took a swing at this in #216

Rigellute:
#216 closes this

PR #216, opened 2020-01-18
Disallow CTRL+d scroll past playlist end

Monkeyanator:
Addresses #192. 

Previously, when using CTRL+d to page downwards on a playlist, there was no guard against incrementing the playlist offset past the end of the playlist, resulting in empty track tables. This commit checks if the track table is empty after attempting a CTRL+d scroll and backtracks to
the previous playlist offset.

From what I can tell, the Spotify API doesn't offer a way to get the total size of a playlist, so we can't just check before making a request whether we're about to scroll past the end (which would be ideal). This means that hitting CTRL+d at the end of a playlist results in two API requests (one to notice we got an empty response, and another to reload with the old limit), and consequently this implementation is a bit on the slow side.

I'm quite new to Rust, so please let me know if I've done anything barbarous :)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -239,7 +239,7 @@ pub struct App {
     pub library: Library,
     pub playlist_offset: u32,
     pub playback_params: PlaybackParams,
-    pub playlist_tracks: Vec<PlaylistTrack>,
+    pub playlist_tracks: Option<Page<PlaylistTrack>>,
     pub playlists: Option<Page<SimplifiedPlaylist>>,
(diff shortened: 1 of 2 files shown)


Monkeyanator:
Cool, didn't realize that field existed! Just updated the PR to use the approach you laid out.

Rigellute:
@all-contributors please add @Monkeyanator for code

Rigellute:
@all-contributors please add @Monkeyanator for code

Issue #213, opened 2020-01-17
Can I play all songs from an artist?


NeverI:
I think this feature is missing from the official app too and I do not understand why :sweat_smile: 

Okay, I can create a playlist with all her/hist album, but it would be awesome just hit a button and start playing all songs without bothering whatever album on its.

Rigellute:
This should be fairly easy for us to implement. We just need to pass in the `artist_uri` to the `start_playback` context parameter. 

We also need to create a new binding to trigger this.

Contributions welcome!

AlexandreSi:
Hi there,
I would be keen on opening my first PR on your repo!
Do you have some indications maybe?

Rigellute:
Closed in #228

PR #228, opened 2020-01-27
Play every known song with e

AlexandreSi:
Hi there,

My first PR to address #213.

I added the ability to play every song for an artist from `artists`, I think it corresponds to the `Artists` tab from `Library`. I did it according to @Rigellute indications.

I am not sure if it should be implemented in another place.

Don't hesitate if you have any comments!

### DIFF
--- a/src/ui/help.rs
+++ b/src/ui/help.rs
@@ -57,7 +57,7 @@ pub fn get_help_docs() -> Vec<Vec<&'static str>> {
             "r",
             "Selected block",
         ],
-        vec!["Play all tracks for artist", "e", "Selected block"],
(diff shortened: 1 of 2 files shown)


src/ui/help.rs
```diff
@@ -57,6 +57,7 @@ pub fn get_help_docs() -> Vec<Vec<&'static str>> {
             "r",
             "Selected block",
         ],
+        vec!["Play all tracks for artist", "e", "Selected block"],
```
Rigellute:
```suggestion
        vec!["Play all tracks for artist", "e", "Library -> Artists"],
```

NeverI:
Awesome, thank you very much!

PR #278, opened 2020-02-02
 Implement 'Made For You' section of Library

amgassert:


src/app.rs
```diff
@@ -1198,4 +1203,54 @@ impl App {
...
+    }
+
+    fn made_for_you_search_and_add(&mut self, search_string: &str) {
+        const SEARCH_LIMIT: u32 = 10;
+        const SPOTIFY_ID: &str = "spotify";
+
+        if let (Some(spotify), Some(user)) = (self.spotify.clone(), self.user.clone()) {
+            let user_country =
+                Country::from_str(&user.country.unwrap_or_else(|| "".to_string())).ok();
+            match spotify.search_playlist(search_string, SEARCH_LIMIT, 0, user_country) {
```
Rigellute: `SEARCH_LIMIT` should be read from `self.large_search_limit`

src/handlers/track_table.rs
```diff
@@ -59,12 +59,10 @@ pub fn handler(key: Key, app: &mut App) {
                             }
                         };
                     }
-                    TrackTableContext::RecommendedTracks => {}
                     TrackTableContext::SavedTracks => {
                         app.get_current_user_saved_tracks_next();
                     }
-                    TrackTableContext::AlbumSearch => {}
-                    TrackTableContext::PlaylistSearch => {}
+                    _ => {}
```
Rigellute: I tend to prefer to explicitly match on enums to let the compiler complain when a new item is added to the enum that is not yet matched. Could you revert this change and also line 93?

src/handlers/track_table.rs
```diff
@@ -123,37 +119,30 @@ fn handle_recommended_tracks(app: &mut App) {
 }
 
 fn jump_to_end(app: &mut App) {
-    match &app.track_table.context {
-        Some(context) => match context {
-            TrackTableContext::MyPlaylists => {
-                if let (Some(playlists), Some(selected_playlist_index)) =
-                    (&app.playlists, &app.selected_playlist_index)
+    if let Some(context) = &app.track_table.context {
+        if let TrackTableContext::MyPlaylists = context {
```
Rigellute: Could this `if let` just become a normal `if context == TrackTableContext::MyPlaylists`?

src/handlers/track_table.rs
```diff
@@ -236,31 +224,43 @@ fn on_enter(app: &mut App) {
                     app.start_playback(context_uri, None, Some(app.track_table.selected_index));
                 };
             }
+            TrackTableContext::MadeForYou => {
+                if let Some(_track) = tracks.get(*selected_index) {
+                    let context_uri = Some(
+                        app.library.made_for_you_playlists.pages[0]
```
Rigellute: Guessing your TODO for pagination is relevant here

src/ui/util.rs
```diff
@@ -61,7 +62,9 @@ pub fn millis_to_minutes(millis: u128) -> String {
 pub fn display_track_progress(progress: u128, track_duration: u32) -> String {
     let duration = millis_to_minutes(u128::from(track_duration));
     let progress_display = millis_to_minutes(progress);
-    let remaining = millis_to_minutes(u128::from(track_duration) - progress);
+
+    let safe_remaining_ms = max(i128::from(track_duration) - progress as i128, 0) as u128;
```
Rigellute: To make this "more safe" you could use `track_duration.checked_sub(progress)` or something to that effect?

src/handlers/track_table.rs
```diff
@@ -94,6 +113,21 @@ pub fn handler(key: Key, app: &mut App) {
                     }
                     TrackTableContext::AlbumSearch => {}
                     TrackTableContext::PlaylistSearch => {}
+                    TrackTableContext::MadeForYou => {
```
amgassert: TODO: review

src/handlers/track_table.rs
```diff
@@ -65,6 +65,25 @@ pub fn handler(key: Key, app: &mut App) {
...
```
amgassert: TODO: review

src/app.rs
```diff
@@ -718,6 +727,26 @@ impl App {
         }
     }
 
+    pub fn get_made_for_you_playlist_tracks(&mut self, playlist_id: String) {
```
amgassert: TODO: review

amgassert:
I've marked the new sections that need review, since I didn't put them in a separate commit. Thanks!

src/handlers/track_table.rs
```diff
@@ -65,6 +65,25 @@ pub fn handler(key: Key, app: &mut App) {
...
+                        let (playlists, selected_playlist_index) =
+                            (&app.library.made_for_you_playlists, &app.made_for_you_index);
+
+                        if let Some(selected_playlist) = playlists.pages[0]
```
Rigellute: Instead of reading the array at index 0, you should use the `get_results` method implemented on the `ScrollableResultPages` struct. The will by default fetch the page in the struct's index.

src/app.rs
```diff
@@ -46,8 +46,8 @@ const DEFAULT_ROUTE: Route = Route {
 
 #[derive(Clone)]
 pub struct ScrollableResultPages<T> {
-    index: usize,
-    pages: Vec<T>,
+    pub index: usize,
+    pub pages: Vec<T>,
```
Rigellute: Does this need to be public? The `impl` below had methods for fetching the data out of `pages`

src/app.rs
```diff
@@ -1198,4 +1227,55 @@ impl App {
...
+                        .items
+                        .iter()
+                        .filter(|playlist| {
+                            playlist.owner.id == SPOTIFY_ID && playlist.name == search_string
+                        })
+                        .map(|playlist| playlist.to_owned())
+                        .collect::<Vec<SimplifiedPlaylist>>();
+
+                    if !self.library.made_for_you_playlists.pages.is_empty() {
+                        self.library.made_for_you_playlists.pages[0]
```
Rigellute: Instead of reading the array at index 0, you should use the `get_results` method implemented on the `ScrollableResultPages` struct. The will by default fetch the page in the struct's index.

src/handlers/track_table.rs
Rigellute: Same here (can use `get_results`)

src/app.rs
```diff
@@ -1262,7 +1269,10 @@ impl App {
                         .collect::<Vec<SimplifiedPlaylist>>();
 
                     if !self.library.made_for_you_playlists.pages.is_empty() {
-                        self.library.made_for_you_playlists.pages[0]
+                        self.library
+                            .made_for_you_playlists
+                            .get_mut_results(None)
```
amgassert: This seems like a better way to handle getting a mutable reference to a page. What do you think?

Rigellute: Looks great 👍

PR #294, opened 2020-02-13
Experimental audio visualiser

Rigellute:
The spotify API has a [audio analysis endpoint](https://developer.spotify.com/documentation/web-api/reference/tracks/get-audio-analysis/) for a given track.

This is an experimental attempt to use that data to "visualise" a track using a bar chart.

### Gif demo
spotify-tui-visualisation

### Image
Screenshot 2020-02-12 at 18 52 32

## How does this work?
This is very basic but the visualisation is based on the "pitch" information for a given "section" of the piece. To make things more animated, in between the section data I create an offset and subtract from the pitch information based on the current beat.

You can now also configure the `tick_rate` of the app. The lower this number the smoother the animations are. But this comes at a cost of higher CPU usage!

## TODO

- [x] Add `v` binding to go to visualisation screen
- [x] Make `tick_rate` configurable via cli arg and config.yml
- [x] Add colors for visualisation to user theme config

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ behavior:
   seek_milliseconds: 5000
   volume_increment: 10
   # The lower the number the higher the "frames per second". You can decrease this number so that the audio visualisation is smoother but this can be expensive!
-  tick_rate: 250
+  tick_rate_milliseconds: 250
 
 keybindings:
   # Key stroke can be used if it only uses two keys:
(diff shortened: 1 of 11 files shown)


echoSayonara:
Great! I love this and look forward to merge this PR.

I have a question about the shown notes in the sample image.
Ten notes are shown in the gif and image. However, an equal temperament that is commonly used contains twelve notes. It seems that A# and H are missing. Or, bar charts only show notes used in a given track?

Rigellute:
@echoSayonara you were right, I wasn't properly displaying all the notes due to the size of the bars. This might still be an issue in smaller terminals but is currently fixed in full screen terminals

image

PR #296, opened 2020-02-14
Modify instruction to install with snap

echoSayonara:
Thanks to contibutions by @popey, @AivreeG, and @hurricanehrndz, we can install `spt` with `snap`.

However, the instruction written in `README` is not updated.
Here, I updated the instruction to install stable version of `spt` with `snap` and add a new instruction to install nightly build of it.


Rigellute:
Thanks @echoSayonara

PR #298, opened 2020-02-16
Display Artist layout when searching an artist url

AlexandreSi:


AlexandreSi:
I am not sure if I should refacto `app.get_artist` to not use the `artist_name` argument and fetch the artist systematically. On one hand, I find it superfluous, on the other hand it can spare an api call.

AlexandreSi:
The next thing to do is to add a key to copy an artist url.

Issue #207, opened 2020-01-10
Extremely slow


gcantieni:
The app barely responds for me. I'll type `?` and it will take a minute or more before anything happens. It seems to work, e.g. connect to Spotify and has all my music, but it is delayed heavily with every action. I'm not sure if this is known, it might be the same bug as #92 but for me it was never "unfrozen".

I got spotifyd and spotify-tui from the AUR. 

Versions:
Operating System: Arch Linux
Kernel: Linux 5.4.8-arch1-1
spotifyd 0.2.20
spotify-tui 0.11.0

EduRenesto:
It's running very slow to me as well. I'm on a very similar setup:

- Arch Linux
- Linux 5.4.7-arch1-1
- spotifyd 0.0.23
- spotify-tui 0.10.0

I'm using Alacritty and zsh.

Rigellute:
Sorry to hear this.

I've just released a new version (0.12.0) which changes the terminal rendering backend (to support Windows).

Are you able to try again?

echoSayonara:
Latest `rspotify` (added `async/await` support) may solve this problem.
https://github.com/ramsayleung/rspotify/issues/68
https://github.com/ramsayleung/rspotify/pull/70

Rigellute:
I'm looking forward to this! It seems `rspotify` is still using the `blocking` variant of `reqwest` so we won't be able to use `async/await` just yet.

I do think this will improve performance.

giorod:
FYI, extremely slow on my system aswell.

Arch Linux
Linux 5.4.7-arch1-1
spotifyd 0.2.24
spotify-tui 0.13.0

Rigellute:
This is looking like an issue with Arch Linux, I don't have an Arch machine to hand to debug this unfortunately.

Timvde:
FWIW: also an Arch user, and spotify-tui has always been snappy for me. So I don't think it's _just_ Arch.

Rayqan:
Also running Arch and I have seen pretty smooth  performance on shorter lists but some "input" lag when scrolling through larger lists like my Liked Songs.

roma41:
Same issue. Thought syslogs would hint me on what the cause of these lags were, but turns out - nothing odd. Also tried switching from Gnome to I3, but no result either. As of now am going to test application via Manjaro, to make sure whether it is specifically Arch Linux only issue.

EduRenesto:
[...]

Issue #92, opened 2019-10-20
Freezes


m-cat:
I had it running for a few songs and when I came back I found the program had frozen. No keypresses worked anymore, and I couldn't even Ctrl+C out.

- OSX 10.14.6 (18G103)
- Terminal 2.9.5 (421.2)
- Fish shell 3.0.2
- GNU Emacs 26.1 (build 1, x86_64-apple-darwin14.5.0, NS appkit-1348.17 Version 10.10.5 (Build 14F2511)) of 2018-05-31

trames57:
This also happening to me.  Just installed it on Arch Linux and if I go away from the app to another workspace (i3) and go back to that workspace, it can freeze. Well actually it is responding a little, but very slowly. It would seem that it may be in some kind of loop and eventually gets some system time to respond to the keyboard. 

I am playing an album, and it is delivering the tracks to the album like it should through the sound.  Eventually I got an API Error "Spotify API reported an error". I hit esc and nothing is happening :(  My album is playing fine though!  OK, after about 4 minutes, it did respond to the esc key. The track time bar and track info at the bottom of the screen has not changed at all. It looks like when it froze, the track had completed (bar is all the way across) and was going to start to show the next track in the album. Weird though that the sound continues as if it not frozen.

OK.. So I had the Spotify web client open at the same time as spotify-tui.  It appears that the web client was playing the tracks during the freeze. So somehow the Spotify web app knows I am playing those tracks (I guess through spotifyd).  I wonder if that was affecting the spotify-tui somehow with the API errors.  I have the web client open now along with spotify-tui and they are coordinated with the same track. They seem to interact with each other and perhaps it is a "racing" issue where the Web client changes tracks before tui or something like that. If I pause in web client, tui pauses and vice-versa.  If I close the web client, tui continues to play.  Perhaps we should not have the web client open along with tui?

Arch Linux is up to date - just updated today (Nov 17, 2019)
Using XFCE terminal in an xwindows session (i3 window manager)
Installed using the arch linux AUR (spotify-tui) and it pulled the latest git release as of 11/17/2019. App built with no errors.
Using spotifyd daemon

trames57:
Just to report back in.  The app froze again playing an album and I did NOT have the web client open. It froze when it went to change to the next track in the album. I was actually watching it LOL. The music continued to play as normal, but the tui client did not change to the next track.

Rigellute:
Thanks for the detailed follow up.

[...]

PR #322, opened 2020-03-05
Create a thread for network events

Rigellute:
This is a huge PR that changes the architecture of the app considerably.

The idea is to create thread that will handle network events, while the main thread will handle the UI rendering loop.

The implication of this change is that the UI thread will disptach (or send) messages to the network thread for it then to execute the async action.

Thus, slow network calls will no longer block the UI thread meaning the app should feel more responsive on slow connections.

I have tested on very slow internet connections and can see these changes will drastically improve both #92 and #207. 

Given the scale of these changes, I encourage anyone to checkout this branch and test!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -418,7 +418,7 @@ name = "failure_derive"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "proc-macro2 1.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 1.0.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "quote 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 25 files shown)


ericonr:
Just started trying it out, thanks for all the hard work! It seems to be much more responsive.

Should I make issues in the issue tracker for this?

When I press `v` to open the audio visualizer, if I press `q` before it loads the visualizer, the quit action is passed to whatever context I'm in, meaning it can even close the application. Testing in the previous version, pressing `q` before the visualizer opened just means that the visualizer will quickly flash on the screen, because it's closed right after.

Perhaps you can announce this update somewhere, so people come try it out? Somewhere like r/unixporn or r/rust.

Rigellute:
Thanks @ericonr.

Please do create an issue for that! 

I think we should navigate _before_ the async action has completed to make it feel more responsive. I also think there will be plenty of other issues caused by the move to full async 😄 .

And absolutely, once this is released I intend to announce on r/rust and r/unixporn!

nahue086:
Amazing. Just tried it out. It works like a charm. Thank you very much!

PR #329, opened 2020-03-10
Mark up saved contents

echoSayonara:
Adding, deleting, following and unfollowing contents is implemented but one can not see a content is saved or follwed from search result view.
In this PR, I implemented marking up a  saved/followed content with `♥` to allow one can see that the content has been already in one's library and rerender search result after adding, deleteing, following, or unfollowing a content.


src/network.rs
```diff
@@ -857,6 +865,37 @@ impl<'a> Network<'a> {
     };
   }
 
+  async fn get_followed_all_artists(&mut self, after: Option<String>) {
+    let mut after = after;
+    loop {
```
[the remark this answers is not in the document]
echoSayonara: I changed to use `check` instead of fetching all albums or artists.
Please check it.

src/network.rs
```diff
@@ -424,7 +432,16 @@ impl<'a> Network<'a> {
     // Run the futures concurrently
     match try_join!(search_track, search_artist, search_album, search_playlist) {
       Ok((track_results, artist_results, album_results, playlist_results)) => {
+        // let artist_ids: Vec<String> = artist_results.artists.items.iter().map(|item| item.id.to_owned()).collect();
```
Rigellute: Are you able to remove this commented code?

src/network.rs
```diff
@@ -876,6 +912,29 @@ impl<'a> Network<'a> {
...
+    if let Some(albums) = &app.search_results.albums {
+      let mut album_ids: Vec<String> = Vec::new();
+      albums.albums.items.iter().for_each(|item| {
+        if let Some(id) = &item.id {
+          album_ids.push(id.to_owned());
+        }
+      });
+      if let Ok(are_follwed) = self
+        .spotify
+        .current_user_saved_albums_contains(&album_ids)
```
Rigellute: Nice 👍

src/network.rs
```diff
@@ -876,6 +912,29 @@ impl<'a> Network<'a> {
...
+        .await
+      {
+        album_ids.iter().enumerate().for_each(|(i, id)| {
+          if are_follwed[i] {
```
Rigellute: Could remove from `saved_album_ids_set` when not followed? Could worth checking if the given album was unfollowed whilst here?

src/network.rs
```diff
@@ -857,6 +874,25 @@ impl<'a> Network<'a> {
...
+    if let Some(artists) = &app.search_results.artists {
+      let artist_ids: Vec<String> = artists
+        .artists
+        .items
+        .iter()
+        .map(|item| item.id.to_owned())
+        .collect();
+      if let Ok(are_follwed) = self.spotify.user_artist_check_follow(&artist_ids).await {
+        artist_ids.iter().enumerate().for_each(|(i, id)| {
+          if are_follwed[i] {
```
Rigellute: Could remove from `followed_artist_ids_set` when not followed?

echoSayonara:
done.

src/network.rs
```diff
@@ -857,6 +865,27 @@ impl<'a> Network<'a> {
     };
   }
 
+  async fn user_artist_check_follow(&mut self) {
+    let mut app = self.app.lock().await;
```
Rigellute: Ah just noticed this!

One problem with this new threaded async architecture is that if we aquire the `lock` _before_ the async action, we can end up blocking the UI rendering loop (since the start of the loop also aquires a `lock`).

This means we must be careful to only the the `lock` _after_ the async action is finished (i.e. when we need to update `app`).

To fix, we must dispatch the data needed for the network request like this

```rust
// From some handler somewhere
 if let Some(artists) = &app.search_results.artists {
      let artist_ids: Vec<String> = artists
        .artists
        .items
        .iter()
        .map(|item| item.id.to_owned())
        .collect();

      app.dispatch(IoEvent::UserArtistFollowCheck(artist_ids));
}

// in network.rs
async fn user_artist_check_follow(&mut self, album_ids: Vec<String>)
```

You can see this pattern in the other handlers.

This is an annoying gotcha that I'd like to solve somehow

echoSayonara: Fixed. However, this change cause one problem.
Just after I searched with one keyword, no albums or artists are marked up even if they are saved or followed and I searched with the same keyword, they are marked up successfully.
Ithink that artsis and album ids were `None` at the first searching because `get_search_result` did not finished at that time.

echoSayonara: In previous commit, id vector is generated after `get_search_result` finished and such problem did not occur. I'm not sure which is better so please check the problem and make decision.

Rigellute:
Thanks for all the work on this @echoSayonara, and sorry for the delay.

I've fixed the last issue you were having in here https://github.com/Rigellute/spotify-tui/pull/342

Closing this in favour of that PR.

PR #342, opened 2020-03-17
Mark up saved contents

Rigellute:
Original PR here https://github.com/Rigellute/spotify-tui/pull/329


Rigellute:
@echoSayonara I've fixed the problem you were facing with search results not showing if they were saved.

You were right, the albums and artists were not resolved at the time we were trying to run the check requests.

To fix, I've needed to dispatch the request from within the `network` thread once the search results are resolved.

PR #388, opened 2020-04-09
Fix crash when there're no artists avaliable

slumber:
Simple check for an out of bounds panic.

I don't really use Spotify, but I was curious to try your application, so after creating new account I've encountered this segfault.

### DIFF
--- a/src/handlers/artists.rs
+++ b/src/handlers/artists.rs
@@ -42,9 +42,11 @@ pub fn handler(key: Key, app: &mut App) {
     }
     Key::Enter => {
       let artists = app.artists.to_owned();
-      let artist = &artists[app.artists_list_index];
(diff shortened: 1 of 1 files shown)


src/handlers/artists.rs
```diff
@@ -42,9 +42,11 @@ pub fn handler(key: Key, app: &mut App) {
     }
     Key::Enter => {
       let artists = app.artists.to_owned();
-      let artist = &artists[app.artists_list_index];
-      app.get_artist(artist.id.clone(), artist.name.clone());
-      app.push_navigation_stack(RouteId::Artist, ActiveBlock::ArtistBlock);
+      if !artists.is_empty() {
```
Rigellute, line 45: Thanks for the contribution @slumber 👍

Rigellute:
@all-contributors please add @slumber for code

PR #409, opened 2020-04-22
Small fixes

TimotheeGerber:
This PR fixes several small things I encountered while using `spotify-tui`, reading the doc or reading the code, like missing spaces or wrong variable name.

I tried to update the API table to better show the work already done. I may have forget some features as I don't use them all.

PR #413, opened 2020-04-24
Fix device selection screen formatting

ClementTsang:
Very minor thing I noticed when setting up - the text looks a bit wonky in the device selection screen.

Currently:
image

I just made it one string instead with proper spacing and added a period, which (IMO) looks nicer:

image


ClementTsang:
Hmm, seems https://github.com/Rigellute/spotify-tui/pull/409 actually already fixes this.

Guess I'll close.

PR #425, opened 2020-04-30
Add more readline shortcuts and make ^U work like readline

fratajczak:


src/handlers/input.rs
```diff
@@ -9,11 +9,37 @@ use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
...
   match key {
+    Key::Ctrl('k') => {
+      app.input.drain(app.input_idx..app.input.len());
+    }
     Key::Ctrl('u') => {
+      app.input.drain(..app.input_idx);
+      app.input_idx = 0;
+      app.input_cursor_position = 0;
+    }
+    Key::Ctrl('c') => {
```
Rigellute, line 20: `Ctrl('c')` is the global exit key of the program (used [here](https://github.com/Rigellute/spotify-tui/blob/master/src/util.rs#L24)).

What other keybind would be appropriate here instead?

fratajczak:
Maybe Ctrl-L (Clear screen in readline) ?

Rigellute:
`Ctrl-l` sounds good 👍

fratajczak:
I just changed it, should be OK now!

Rigellute:
@all-contributors please add @fratajczak for code

PR #442, opened 2020-05-06
moved pagination to start of help-file

### LINKED ISSUE #439: help menu has pagination key too far down (hen/egg problem)
when looking at the help menu, the key for pagination (ctrl-d) is only mentioned so far down that on regular terminals, you won't see it on the first page. Meaning that if you don't know that ctrl-d is for pagination and head to help to find out all the keys, you'll never know that there's more than one page of content there (or anywhere else for that matter). Trying space or pg-down (which I did) obviously doesn't help. Only reading some (as far as I could tell) unrelated issue about playlists here caused me to realize that ctrl-d was for pagination (since it's not really discoverable - I mean UNIX convention would dictate a different meaning to ctrl-D, right?).

So .. put it in the first line of the help, maybe?

mhellwig:
like this?
resolves #439

### DIFF
--- a/src/ui/help.rs
+++ b/src/ui/help.rs
@@ -1,5 +1,13 @@
 pub fn get_help_docs() -> Vec<Vec<&'static str>> {
   vec![
+    vec!["Scroll down to next result page", "<Ctrl+d>", "Pagination"],
+    vec![
+      "Scroll up to previous result page",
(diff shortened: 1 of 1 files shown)


mhellwig:
resolves #439

Rigellute:
@all-contributors please add @mhellwig for doc

PR #445, opened 2020-05-08
Add support for saving an album and following an artist in artist view

GrudgeDuck:
Currently, album save and artist follow are enabled in search result only.
In this PR, I added support for album save/delete and artist follow/unfollow in artist view.

### DIFF
--- a/src/handlers/artist.rs
+++ b/src/handlers/artist.rs
@@ -1,5 +1,5 @@
 use super::common_key_events;
-use crate::app::{App, ArtistBlock, RecommendationsContext, TrackTableContext};
+use crate::app::{ActiveBlock, App, ArtistBlock, RecommendationsContext, TrackTableContext};
 use crate::event::Key;
 use crate::network::IoEvent;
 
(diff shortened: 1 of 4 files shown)


src/handlers/artist.rs
```diff
@@ -296,6 +296,16 @@ pub fn handler(key: Key, app: &mut App) {
           handle_recommend_event_on_selected_block(app);
         }
       }
+      Key::Char('w') => match artist.artist_selected_block {
+        ArtistBlock::Albums => app.current_user_saved_album_add(ActiveBlock::ArtistBlock),
+        ArtistBlock::RelatedArtists => app.user_follow_artists(ActiveBlock::ArtistBlock),
+        _ => (),
+      },
+      Key::Char('D') => match artist.artist_selected_block {
+        ArtistBlock::Albums => app.current_user_saved_album_delete(ActiveBlock::ArtistBlock),
```
Rigellute, line 305: As a side note, these destructive actions should now be handled after a confirmation dialog.

This can be done in another PR though.

PR #446, opened 2020-05-10
Move "Pagination" commands to top of help menu

sheelc:
Allows for someone reading the help menu to know that they could paginate
through the help menu. [#439]


sheelc:
Whoops, didn't see that that was gotten to already in #442 -- closing in favor of the PR from the original author of the issue!

PR #472, opened 2020-05-21
Change currently_playing indicator

oliver-daniel:
Since we're already using the Geometric block for the heart, it just made sense to use something a little more elegant for the current indicator. :) Let me know if this causes any issues.


Rigellute:
@all-contributors please add @oliver-daniel for code

Rigellute:
@all-contributors please add @oliver-daniel for code

PR #502, opened 2020-06-20
Simple change of layout 

mucinoab:
`draw_input_and_help_box` moved to `draw_user_block` so the search and help bar is reduced and the main panel is adjusted to fill that space. The constraints of size of both functions were tweaked as well to better fit the new layout. 

I feel this new layout does a better job at taking advantage of space as the strings we search are usually not that long and after some use of the client the help bar becomes more or less irrelevant. 

Here some screenshots 

Fullscreen in a 1080p monitor 
Screenshot_full

Halfscreen in a 1080p monitor 
Screenshot_half

Current layout for comparison 
old

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -186,7 +186,7 @@ where
           Constraint::Min(1),
           Constraint::Length(6),
         ]
-        .as_ref()
+        .as_ref(),
       )
       .margin(margin)
       .split(f.size());
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -20,7 +20,7 @@ use tui::{
 use util::{
   create_artist_string, display_track_progress, get_artist_highlight_state, get_color,
   get_percentage_width, get_search_results_highlight_state, get_track_progress_percentage,
-  millis_to_minutes,
+  millis_to_minutes, SMALL_TERMINAL_WIDTH,
 };
 
 pub enum TableId {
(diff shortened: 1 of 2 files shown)


mucinoab:
@Rigellute No, the new layout I propose will display always, no matter the size of the screen.

Rigellute:
So sorry for the late response @mucinoab. 

What I mean to say is that we should pick some breakpoint at which your new layout kicks in.

For example, we might say on larger screens (width = 800px) your new layout is active. And the old layout is active for smaller screens.

Hope this makes sense?

If you don't know what I mean, I could merge this in to a side branch and have a look myself?

mucinoab:
@Rigellute 
I see what you mean, that seems a good idea. I am not very familiar with the tui crate but let me give it a go, if any trouble comes out I will let you know so you can go ahead with the branch. 
Thanks, don't worry about the late response.

mucinoab:
@Rigellute I didn't find a nice way of doing the dynamic change, all my tries were mess, buggy or didn't work at all. :(
Could you give it a go? I think is a nice feature.

OrangeFran:
I added responsiveness and openen a pull request on the fork of @mucinoab . The point of change is defined in the `src/ui/util.rs` as `SMALL_TERMINAL_WIDTH` and is currently 150. I used `app.size.width` to check for the width of the terminal and changed the layouts accordingly.

Rigellute (approved):
Excellent work @mucinoab @OrangeFran 👏

I just tested this and it looks great!

Thank you both 🙏

Rigellute:
@all-contributors please add @mucinoab and @OrangeFran for code

Rigellute:
@all-contributors please add @OrangeFran for code

Rigellute:
@all-contributors please add @OrangeFran for code

PR #509, opened 2020-06-24
Add option to vertically center or decenter basic view playbar

### LINKED ISSUE #508: Basic view sacrifices content before margin in small terminals
ezgif com-video-to-gif

In a terminal with a small text size to terminal size ratio, the content of the basic view window is removed before the vertical margin/padding

fezboy:
doesn't quite resolve #508, but adds a workable option for usecases with a small terminal height (i.e. playbar across bottom of screen in tiling wm.
Center basic view on (default, old):
image
image

Center basic view off:
image
image


Rigellute:
Thanks for this @The-Bob. 

I can't help but think there must be a better way to center a view. My solution was not good at all.

Perhaps there's an example of proper center alignment in the [`tui-rs`](https://github.com/fdehau/tui-rs)

OrangeFran:
Hi there.

TL;DR Simpler code by manually calculating space. If ok, should I create a new pr here or on the linked fork?

I looked into this and it seemed to me that the simplest way to center the widget would be to calculate the space manually. The changes and calculations are very simple. Something like this:
``` rust
let space = (app.size.height as i16 - BASIC_VIEW_HEIGHT as i16) / 2;
```
BASIC_VIEW_HEIGHT here is 7. Unfortunately there are these integer conversions because the result could actually be negative and thus the code could panic. The signed integers solve this but I have to check afterwards it it's positive and convert back to `u16`.

If that solves the issue, should I create a pull request on the fork of @The-Bob or create a new pr.

fezboy:
@OrangeFran you should make a separate PR for it, because mine's just a dirty workaround.

I tested your branch and it works perfectly, thank you so much for working that out!

OrangeFran:
Glad it works and you like it.
I'll create a PR.

Issue #495, opened 2020-06-11
Cannot save songs


Babbie:
Whenever I try to press the hotkey for saving the currently highlighted song in the table, nothing happens. I have not altered the key in my config.yml, so ```s``` should still be assigned to it. My config.yml, just to be sure:
```
# Sample config file

# The theme colours can be an rgb string of the form "255, 255, 255" or a string that references the colours from your terminal theme: Reset, Black, Red, Green, Yellow, Blue, Magenta, Cyan, Gray, DarkGray, LightRed, LightGreen, LightYellow, LightBlue, LightMagenta, LightCyan, White.
theme:
  active: Cyan
  banner: LightCyan
  error_border: Red
  error_text: LightRed
  hint: Yellow
  hovered: Magenta
  inactive: Gray
  playbar_background: Black
  playbar_progress: LightCyan
  playbar_text: White
  selected: LightCyan
  text: White

behavior:
  seek_milliseconds: 5000
  volume_increment: 10
  # The lower the number the higher the "frames per second". You can decrease this number so that the audio visualisation is smoother but this can be expensive!
  tick_rate_milliseconds: 250

keybindings:
  # Key stroke can be used if it only uses two keys:
  # ctrl-q works,
  # ctrl-alt-q doesn't.
  back: "esc"

  jump_to_album: "a"

  # Shift modifiers use a capital letter (also applies with other modifier keys
  # like ctrl-A)
  jump_to_artist_album: "A"

  manage_devices: "d"
  decrease_volume: "-"
  increase_volume: "+"
  toggle_playback: " "
  seek_backwards: ","
  seek_forwards: "."
  next_track: ">"
  previous_track: "<"
  copy_song_url: "c"
  copy_album_url: "C"
  help: "?"
  repeat: "r"
  search: "/"
  audio_analysis: "v"
```

Rigellute:
Hmm... strange. Thanks for reporting - will look into it.

jackson15j:
With the default keymap I can not save songs from a Playlist. Here are my reproduction steps.

* `$ spt`.
* `Library <return> Made For You <return> Discover Weekly <return>`.
* `s` on highlighted song does not save the song to my `Liked Songs`.
* `<return>` to play a song and then `s` does save the song to my `Liked Songs`.
* `a` to jump to album of currently playing song and then `s` on **any** song in that album will save that respective song to my `Liked Songs`.

Expected `s` to work in all 3 scenarios.

[...]

PR #525, opened 2020-07-10
Handle track saving in non-album contexts (eg. playlist/Made for you).

jackson15j:
Saving of tracks used to only work in album view (handled by:
`src.handlers.album_tracks.handle_save_album_event`).

Add some rough code to get track saving working in other cases. Manually tested
`Playlist` and `Made for you` views.

Fixes Rigellute/spotify-tui#495


jackson15j:
Apologies if the code is dirty. First time looking at Rust. Let me know what additional work you'd like me to do to progress this further.

src/handlers/track_table.rs
```diff
@@ -252,6 +253,16 @@ fn play_random_song(app: &mut App) {
   };
 }
 
+fn handle_save_track_event(app: &mut App) {
+  let (selected_index, tracks) = (&app.track_table.selected_index, &app.track_table.tracks);
+  if let Some(track) = tracks.get(*selected_index) {
+    if let Some(id) = &track.id {
+      let _id = id.to_string();
+      app.dispatch(IoEvent::ToggleSaveTrack(_id));
```
Rigellute, line 261: The underscore prefix (`_id`) is usually used to convey that the variable is not used.

Instead you could do this?

```suggestion
      app.dispatch(IoEvent::ToggleSaveTrack(id.to_string()));
```

jackson15j: Good to know the underscore means unused (Bad habit from other languages, where the underscore prefix is for a private/local variable).

Your suggestion is what I had in the original commit, which broke the build checks.

Rigellute: Ah okay cool, that's all good then!

So if you could just remove the underscore then we are good to go!

jackson15j: Fixed up and also merged in latest master changes. Works locally. So hoping all Github checks still pass.

Rigellute:
Thanks for your contribution @jackson15j. 

As a side note, it is typically preferable to use `git rebase` to sync a branch with master, that way we avoid the merge commit we see above. I need to create a contributors guide detailing this, so no worries for now!

@all-contributors please add @jackson15j for code.

PR #531, opened 2020-07-21
Load playlist listing greedily

### LINKED ISSUE #116: Playlist section not complete
So the playlist section is missing playlists. From the looks of it the playlist api only returns 20 playlists by default. Adding limit=50 to the query will increase it to 50 however in my case I have 70 playlists. In which case paging will need to be implemented. Overwise great job I really enjoy it.

jackkinsey:
Fixes #116. 

This change defies the pagination model the rest of the application seems to be using, requesting pages of playlists until all playlists have been loaded. (It's also slightly hacky in doing so.) 

I totally understand if you don't want to go this route with the app, but I figured I might as well show up with some working code to get the conversation started. (Obligatory note about being a Rust beginner/apologies if my code isn't idiomatic.)

Thanks for making this! It's super cool, I look forward to using it more.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -1093,9 +1093,33 @@ impl<'a> Network<'a> {
       .await;
 
     match playlists {
-      Ok(p) => {
+      Ok(mut page) => {
+        let playlists_page_count = page.total / self.large_search_limit
+          + if page.total % self.large_search_limit > 0 {
+            1
+          } else {
+            0
+          };
+
+        let all_playlists = &mut page.items;
+
+        for i in 1..playlists_page_count {
+          let playlists = self
+            .spotify
+            .current_user_playlists(self.large_search_limit, self.large_search_limit * i)
+            .await;
+
+          match playlists {
+            Ok(p) => {
+              all_playlists.extend(p.items);
+            }
+            Err(e) => {
+              self.handle_error(anyhow!(e)).await;
+            }
+          }
+        }
         let mut app = self.app.lock().await;
-        app.playlists = Some(p);
+        app.playlists = Some(page);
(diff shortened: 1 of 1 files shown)


Rigellute:
Thanks for taking a crack at this @jackkinsey.

I'd prefer us to not greedily load all tracks in a playlist.

Pagination is a tricky one to solve and that is part of the reason not all views implement it currently. 

I need to sit down and have a rethink to hopefully find a more generic solution.

Anyway, appreciate your effort here. And for what it's worth your code looks good!

jackkinsey:
Just to be clear: this change doesn't load all the *tracks in a playlist*, it loads all the *playlists in the playlist listing*, in the sidebar.

Rigellute:
Thanks for the clarification @jackkinsey. 

I meant to say I want us to avoid greedy loading of any paginated response.

Anyway, thanks again for your efforts. 

I will close this and try to find a better generic solution for handling pagination.

jradek:
Hi,
I've version 0.21.0 on MacOS and the problem is somehow still there.
Only a fraction of all my playlists are shown and ctrl-d / ctrl-u don't work. j/k just cycle through the first result page.
Could it be a regression?

JR

j-hui:
@Rigellute Could you explain more about your opposition to this particular workaround? This might not be a generic solution that solves the issue of pagination once and for all, but at least it helps us see playlists we currently simply cannot see (:

j-hui:
@jradek I think the `<C-d>` and `<C-u>` bindings are specifically used cycling through pages of songs _within a playlist_ (rather than cycling through pages of the _list of playlists_). This PR specifically deals with the list of playlists.

PR #537, opened 2020-07-24
Upgrade rspotify to 0.10.0 (Podcast support)

Rigellute:
The latest version of [`rspotify`](https://github.com/ramsayleung/rspotify/blob/master/CHANGELOG.md#010-20200701) brings support for podcasts! 

spotify-tui itself still needs to implement searching and playing podcasts, but this PR will enable its development.

This upgrade introduced a number of breaking changes, but thanks to the amazing type system and the Rust compiler, it was fairly easy to just follow the errors for everything to Just Work!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1446,9 +1446,9 @@ dependencies = [
 
 [[package]]
 name = "rspotify"
-version = "0.9.0"
+version = "0.10.0"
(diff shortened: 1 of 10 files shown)


sputnick1124:
Yay! didn't realize `rspotify` had gotten podcasts figured out! This is great news!

Issue #538, opened 2020-07-24
UI doesn't indicate if album is saved


Rigellute:
Fix is to show the heart icon in the album table if the album is liked.

onielfa:
Hello,

I have opened a new PR ([#540](https://github.com/Rigellute/spotify-tui/pull/540)) to add the heart icon to the saved albums list. I'm very new to the project and maybe I am getting something wrong here. I have looked to the places where the albums are missing the heart icon and the Saved Album was the only one I found. I didn't check if the albums are in the `saved_album_ids_set` as per definition any Saved Album is liked, if I get this right.

onielfa:
@Rigellute Can we close this issue?

Rigellute:
Yes! Thanks for the reminder @onielfa

PR #540, opened 2020-07-24
Show ♥ next to album name in saved list

onielfa:
In order to solve [#538 ](https://github.com/Rigellute/spotify-tui/issues/538)


Rigellute:
@all-contributors please add @onielfa for code

PR #545, opened 2020-07-28
Document Nix support

jwijenbergh:
`spotify-tui` has been in the stable channel for Nix/NixOS for some time now. Would be a really good addition to the installation section as the Nix package manager can be used on MacOS and virtually any linux distro.


Rigellute:
@all-contributors please add @jwijenbergh for doc

PR #550, opened 2020-08-09
Add handling Home and End buttons in user input

murlakatamenka:
Cause why not.


Rigellute:
@all-contributors please add @murlakatamenka for code

PR #563, opened 2020-08-27
Add basic playbar support for podcasts

sputnick1124:
Small-ish PR to get basic podcast support kickstarted. I plan on adding more support for searching and following podcasts, but figured I'd tackle the low hanging fruit first.

Things added:
* playback progress bar
* seek forward/backward
* copy episode/show uri

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,1 +1,2 @@
 tab_spaces=2
+edition = "2018"
--- a/src/app.rs
+++ b/src/app.rs
@@ -426,15 +426,15 @@ impl App {
         .as_millis()
         + u128::from(*progress_ms);
 
-      match item {
-        PlayingItem::Track(track) => {
-          if elapsed < u128::from(track.duration_ms) {
-            self.song_progress_ms = elapsed;
-          } else {
-            self.song_progress_ms = track.duration_ms.into();
-          }
-        }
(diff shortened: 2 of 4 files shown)


rustfmt.toml
```diff
@@ -1 +1,2 @@
 tab_spaces=2
+edition = "2018"
```
sputnick1124, line 2: Not sure if there is another way to allow `rust-fmt` deal with 2018 (read: `async` code), but this shouldn't hurt anything, right?

sputnick1124:
@Rigellute Any update on getting this approved? I'm making good headway on full podcast support over on https://github.com/sputnick1124/spotify-tui/tree/podcast-support-search and I'm almost ready to submit a PR for it, but it builds off of these commits, so it should definitely wait until this goes through.

Rigellute (approved):
So sorry for the long delay @sputnick1124! 

Your changes look great, as always. Thank you 🙏

rustfmt.toml
sputnick1124: Oh ok, that makes sense. I wonder what I might have done to my machine that made it complain for me...

sputnick1124:
No worries! I didn't mean to be pushy! Thanks again for starting and maintaining this whole thing!

Issue #490, opened 2020-06-06
Change current song progress color


tysaya123:
It would be neat if you could change the color of the text that shows when the progress bar covers it without changing the background. My motivation is that I enjoy the black background but the default color can be hard to read on dark progress bar colors.

<img width="1423" alt="Screen Shot 2020-06-06 at 11 50 41 AM" src="https://user-images.githubusercontent.com/16331012/83952212-14b8d780-a7ec-11ea-8aee-2d9753172f1f.png">

macguirerintoul:
On a similar note, the bounding box for the text seems to invert the background. It would be nice if only the text would change colour rather than the background as well.


Before passing:
<img width="1105" alt="image" src="https://user-images.githubusercontent.com/18323154/83956083-fe704300-a80e-11ea-98e0-cbffd2fffb15.png">

After passing:
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/18323154/83956061-c832c380-a80e-11ea-95f3-621e417f57ee.png">

tysaya123:
How were you able to change just the color of the bounding box?

macguirerintoul:
I wasn't able to, I think it just uses `playbar_background` and `playbar_progress` depending on whether it's passed or not

Rigellute:
The should be doable! We just need to add another item to the theme config and use that color in the playbar progress

Peterkmoss:
As of right now tui doesn't support multiple font colors for widgets / elements, so this would require a rework of how the playbar progress is presented.

@macguirerintoul  I can't seem to reproduce that behavior, but I think that's an issue for itself and not this one specifically.

PR #564, opened 2020-08-27
Issue #490 updated tui to 0.10.0 and added playbar_progress_text option

Peterkmoss:
This is from issue #490 

There doesn't seem to be support for changing the color of the text when the bar is overlapping to anything else though. This is hardcoded into tui to use the background of the gauge...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -770,15 +770,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "itertools"
-version = "0.9.0"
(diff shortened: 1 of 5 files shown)


Rigellute:
Nice work @Peterkmoss! I had one go at this upgrade a few weeks ago and never finished. So really appreciate your help here 👏.

Just checked it out and all looks great - noticed one issue with the audio analysis view which I've fixed and commited to this PR.

Once checks pass, I'll merge!

Issue #304, opened 2020-02-20
Add to Queue command


mikeboiko:
Would it be possible to add a command for "Add to Queue"?
I use this feature all the time on my Android phone.
I love spotify-tui, thanks for allowing me to live in my terminal even more!

pedrohva:
I'm pretty interested in this feature as well but [rspotify](https://docs.rs/rspotify/0.8.0/rspotify/spotify/client/struct.Spotify.html) doesn't seem to have any options to add a song to a queue. I can't seem to find anything on the Spotify Web API Reference either.

mikeboiko:
Yea you're right, everyone is still waiting on Spotify to add the API endpoint:
https://github.com/spotify/web-api/issues/462

Rigellute:
I'd love this feature too

Bios-Marcel:
Doesn't look like spotify gives a damn about this. Alternatively a local queue could be added.

Something like "OnSongEnd: if queue.size > 0; play(queue.pop)". The queue could be completly unrelated to playing playlists, albums or radios, just the like the official queue. I think this would be a great compromise.

Rigellute:
Ah cool! Looks like this is a new endpoint not yet covered by `rspotify`, I will look into adding it there so we can start supporting queues!

austenmyers:
Any updates on the queuing feature?

Rigellute:
Looks like `rspotify` has added this endpoint as of yesterday 🎉  https://github.com/ramsayleung/rspotify/commits/master

Once `rspotify` has released we can start working on it 👍

tonitch:
it seems implemented in rspotify now. I really wanted to try implementing it but... 
I neved did rust before :'(
any update for this tho?

Rigellute:
Thanks to @BeneCollyridam this is now possible!

mikeboiko:
Awesome, good work everybody!

PR #567, opened 2020-09-01
Add possibility to queue a song

BeneCollyridam:
Using the new method in rspotify, we can now queue a song.
I have set the deafult to 'z', but it should probably be changed before the code
is merged.

This is in reference to issue #304 

Please let me know if there are better ways to do this, then I can improve the
code

### DIFF
--- a/src/user_config.rs
+++ b/src/user_config.rs
@@ -165,6 +165,7 @@ pub struct KeyBindingsString {
   copy_album_url: Option<String>,
   audio_analysis: Option<String>,
   basic_view: Option<String>,
+  add_item_to_queue: Option<String>,
 }
 
 #[derive(Clone)]
@@ -320,6 +321,7 @@ impl UserConfig {
     to_keys!(copy_album_url);
     to_keys!(audio_analysis);
     to_keys!(basic_view);
+    to_keys!(add_item_to_queue);
 
     Ok(())
   }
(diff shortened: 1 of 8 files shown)


BeneCollyridam:
I am uncertain, what it should say in the help page, but maybe someone else can add that?

src/user_config.rs
```diff
@@ -249,6 +250,7 @@ impl UserConfig {
         copy_album_url: Key::Char('C'),
         audio_analysis: Key::Char('v'),
         basic_view: Key::Char('B'),
+        add_item_to_queue: Key::Char('z'),
```
Rigellute, line 253: I think you'll also want to add `add_item_to_queue` in a few other places in this file. As an example, look at the `basic_view` prop.

Rigellute (commented):
Great work @BeneCollyridam - this looks great! 🎉. This has been a long requested feature.

I have only one comment on adding your new `add_item_to_queue` prop to the user config.

Rigellute:
And you could add something like this to help.rs

```rs
 vec!["Add track to queue", "z", "Hovered over track"],
```

src/user_config.rs
BeneCollyridam: I believe I have fix this in eb6416d

BeneCollyridam:
I added your (@Rigellute) recommended help text. Thanks!

Rigellute (approved):
Awesome 🎉!

Rigellute:
@all-contributors please add @BeneCollyridam for code

PR #580, opened 2020-09-19
Don't add analysis view to stack if already in it.

ericonr:
Before this commit, if a user pressed `v` multiple times, they would
have to press `q` the same number of times to get to the previous view.

This commit makes it so that an analysis view entry is only pushed into
the navigation stack if the current view isn't an analysis one.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -891,9 +891,11 @@ impl App {
     {
       match item {
         PlayingItem::Track(track) => {
-          let uri = track.uri.clone();
-          self.dispatch(IoEvent::GetAudioAnalysis(uri));
-          self.push_navigation_stack(RouteId::Analysis, ActiveBlock::Analysis);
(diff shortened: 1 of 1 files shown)


ericonr:
I think this might be an issue with most views, but it would be quite complex to add it for the others, such as album and artist ones. So I still think it makes sense to add it for this specific one.

Rigellute (commented):
Nice one @ericonr - I think you're right to only fix this for audio analysis at the moment.

Thank you for your contribution.

Rigellute:
@all-contributors please add @ericonr for code

ericonr:
@Rigellute I was thinking of checking if analysis was the current view in the function that pushes it into the stack, but to make that work we'd probably require a bigger design change. I am interested in trying to look into this to make it more generalized. Perhaps return an error from the function that pushes into the stack if the current view is the same? And use that as a conditional for the other actions?

PR #566, opened 2020-08-31
Add next/prev page, jump to start/end to user config

Toaster192:
resolves #482 
@AndydeCleyre, @vidurb might want to know

This allows the 4 keybinds to be bound to something else (something that doesn't colide with people's tmux keybinds and such)

PR #581, opened 2020-09-19
Full Podcast support

sputnick1124:
Fully functional as is, but still missing some features that I think should be implemented before this is considered complete. Here's the status on this thus far. Open to cutting this PR here and opening new PRs for remaining tasks. Otherwise, I plan on getting rid of the draft status when the following are complete:

- [x] Podcast shows search results
* ~~Library podcast view~~
- [x] Podcast show episode table
  - [x] Show current progress in episode in duration field
  - [x] Show completed episodes (use :heavy_check_mark: to signify played status)
  * ~~Display episode description with new hotkey (`CTRL-]`?)~~
  *~~Paged episode results~~
  - [x] Jump to start/end of list. __Note__: will need to be revisited if/when https://github.com/Rigellute/spotify-tui/pull/566 is merged
  - [x] Sort by date ~~/ filter by FullyPlayed status~~
* ~~Follow/Unfollow podcasts (from episode table, show search view, and playbar context)~~

UPDATE: This PR is getting to be pretty big already. I think the other features can be implemented piecemeal in smaller PRs to follow. Some of these are fairly simple and could be good opportunities for newcomers to get their feet wet in the project.

tty
*Note that the above GIF shows the progress bar from https://github.com/Rigellute/spotify-tui/pull/563*


Rigellute:
Wow! Amazing work @sputnick1124 - I will get to reviewing this as soon as I can.

sputnick1124:
Awesome! I wasn't sure where to put the search results without cluttering up the page, while also making them easy to get to, so if the layout seems bad to people, we can figure out a better way. I think tui-rs has tab support, so we could leverage that perhaps. I'm open to any changes that we think should be made to either the UI or implementation.

src/handlers/episode_table.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+    k if common_key_events::low_event(k) => {
+      let next_index = common_key_events::on_low_press_handler(&app.episode_table.episodes);
+      app.episode_table.selected_index = next_index;
+    }
+    Key::Enter => {
+      on_enter(app);
+    }
+    // Scroll down
+    Key::Ctrl('d') => {}
```
Rigellute, line 41: Would be great if the pagination keys referenced the new keys from the user config (` app.user_config.keys.next_page` as implemented here https://github.com/Rigellute/spotify-tui/pull/566).

Not a blocker for this PR though since there is no pagination yet.

src/handlers/episode_table.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+fn toggle_sort_by_date(app: &mut App) {
+  let selected_id = app
+    .episode_table
+    .episodes
+    .get(app.episode_table.selected_index)
+    .map(|e| e.id.clone());
+  app.episode_table.episodes.reverse();
+  app.episode_table.reversed ^= true;
```
Rigellute, line 85: Not seen this `^=` syntax before - can't seem to find reference to it either 🤷‍♂️.

Would you mind explaining what this does?

src/ui/mod.rs
```diff
@@ -1548,6 +1672,20 @@ fn draw_table<B>(
...
+              formatted_row[name_idx] = format!("▶ {}", &formatted_row[name_idx]);
+              style = Style::default()
+                .fg(app.user_config.theme.active)
+                .add_modifier(Modifier::BOLD);
+            }
+          }
+        }
+      }
       _ => {}
     }
```
Rigellute, line 1691: Might be nice to pull in the users podcasts library?

image

This can be done later though if you'd rather not add any more to this PR - I'll leave that up to you.

src/ui/mod.rs
```diff
@@ -1265,6 +1302,93 @@ where
...
+    .episodes
+    .iter()
+    .map(|episode| {
+      let (played_str, time_str) = match episode.resume_point {
+        Some(ResumePoint {
+          fully_played,
+          resume_position_ms,
+        }) => (
+          if fully_played {
+            " ✔".to_owned()
```
Rigellute, line 1354: Very nice touch 👍

Rigellute (commented):
Really impressive work @sputnick1124, this looks great.

I've just got one question about unfamiliar syntax, but we should be good to go.

src/handlers/episode_table.rs
sputnick1124: Sure thing. `x ^= 1` means `x equals x XOR 1`, so it ends up being a convenient way to toggle a boolean value back and forth since `0 ^ 1 = 1` and `1 ^ 1 = 0`. This has been implemented on the core `bool` type ([docs](https://doc.rust-lang.org/std/primitive.bool.html#impl-BitXorAssign%3Cbool%3E)).

src/handlers/episode_table.rs
sputnick1124: Good catch. I agree that this would probably get fixed when paging gets implemented for this table, but I'll pop that in now just so it doesn't get overlooked.

src/ui/mod.rs
sputnick1124: Yeah I felt like that this PR was getting to be a bit large and I didn't want to overwhelm it with changes. I was planning on cutting a new PR with library podcast support shortly following this merge, if that's alright.

src/handlers/episode_table.rs
Rigellute: Ah I see - thank you 👍

src/handlers/episode_table.rs
sputnick1124: Should be fixed now.

Rigellute:
👏

Rigellute:
@all-contributors please add @sputnick1124 for code, bug, maintenance, question, doc

PR #583, opened 2020-09-21
Add user-configurable header styling

n-ivanov:
- Header text in panes is not affected by the color value set for 'text' in config.yml
    - Results in headers always written in white, which causes clash with non-white themes
- Add new theme 'header' setting and use it to style the header text

pre-fix

post-fix


Rigellute (approved):
Nice one 👍. Thank you

Rigellute:
@all-contributors please add @n-ivanov for code

PR #594, opened 2020-09-28
Updated README with correct brew command

Kryan90:
spotify-tui is available as an official tap
Trying to install with the current instructions gives an error
`Error: No available formula or cask with the name "Rigellute/tap/spotify-tui"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ The binary executable is `spt`.
 For both macOS and Linux
 
 ```bash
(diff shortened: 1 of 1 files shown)


Rigellute (approved):
Wow! I did not know that spotify-tui made its way into homebrew core. Thanks for updating this.

Rigellute:
@all-contributors please add @Kryan90 for doc

Issue #527, opened 2020-07-14
Query amount dependent on terminal size?


realmain:
It looks like the amount of information queried is dependent on the size of the terminal? I've tested it on both of my computers with konsole and termite.

This is the terminal size being small:

![](https://i.imgur.com/squKTAh.png)

After I expand the size of my terminal, it shows that it's missing a lot of songs from the playlist, and the playlists section is missing a lot of playlists:

![](https://i.imgur.com/IG11zQQ.png)

If I start `spt` with the same expanded size terminal above, this is what it looks like:

![](https://i.imgur.com/9QvUEW9.png)

There are now more songs and playlists, but it is still missing a lot of them.

AivreeG:
As far as I know, this is expected behavior based on the way `spt` handles pagination. Ctrl+D and Ctrl+U move the viewable section down and up respectively. The displayed height is dependent on the terminal size, naturally. An alternative issue title may be More Dynamic List Rendering for Changing Terminal Sizes.

realmain:
@AivreeG , even the pagination isn't handled well, causing a lot of missing music

The only place I can actually use Ctrl+D & Ctrl+U is the Playlists -> Songs section:

![](https://i.imgur.com/1VMDcTd.png)

If I search for an artist, and go to the songs section, it does not work:

![](https://i.imgur.com/vlfTTBO.png)

This album has 25 songs, but it is only showing 15 songs. Arrow keys will just cycle through 1-15, and Ctrl+D / U does not do anything here:

![](https://i.imgur.com/xOMdfQt.png)

AivreeG:
In that case, I agree. I primarily checked in the saved Playlists and Albums section. I've found a long saved album can be simply arrowed through. This seems to be an issue with search queries specifically, then

Rigellute:
@AivreeG is correct that terminal size defines the amount of data fetched.

The second issue is that pagination is not implemented in every view yet (this is on the [roadmap](https://github.com/Rigellute/spotify-tui#high-level-requirements-yet-to-be-implemented))

Hugo01:
I wonder if there's a way to get rid of pagination and query the entire list at once 
That would "solve" the problem in the "liked songs" sections where you can only listen to songs on that particular page

realmain:
[...]

PR #604, opened 2020-10-04
Implement next/previous page behavior for the Artists table

### LINKED ISSUE #603: Artists pagination is not implemented
First off, great work on this. I was having some memory usage issues on my old laptop and this project shaved off an entire gigabyte of memory usage that the official Spotify desktop client was hogging!

# Steps to reproduce
* Go to the `Artists` page from the `Library` section.
* Try to use `<ctrl+d>`/`<ctrl+u>` to paginate the `Artists` page.
* Note that nothing happens.

Looking at the code this seems to be because we have no handlers for the next and previous page actions. I'll have a PR up for this soon, but I thought that I should probably file the issue first.

Utagai:
This closes #603 which I just filed.

I'm new to Rust, so any feedback is welcome!

Lastly, I think pagination behavior may be in want of a re-work. It seems like we're missing pagination for a bunch of views and I imagine their implementations will all share quite a bit in common. I didn't take on the larger task of a big refactor but it may be something we should consider.

Here's a demo gif (ignore color issues, it's my first time trying to use `terminalizer` to record my terminal).
artists-pagination-demo

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -676,6 +676,52 @@ impl App {
     ));
   }
 
+  pub fn set_saved_artists_to_table(&mut self, saved_artists_page: &CursorBasedPage<FullArtist>) {
+    self.dispatch(IoEvent::SetArtistsToTable(
+      saved_artists_page
+        .items
+        .clone()
+        .into_iter()
+        .collect::<Vec<FullArtist>>(),
+    ))
+  }
+
+  pub fn get_current_user_saved_artists_next(&mut self) {
+    match self
+      .library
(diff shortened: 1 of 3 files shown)


Rigellute (approved):
Good work @Utagai. 

Indeed you're right, pagination is a problematic area of spotify-tui right now.

See the discussion here for a more unified solution https://github.com/Rigellute/spotify-tui/issues/527#issuecomment-700385162

Rigellute:
@all-contributors please add @Utagai for code

PR #612, opened 2020-10-10
Update the saved albums cache set when getting an artist

### LINKED ISSUE #611: Saved albums are not recognized when an artist's page is loaded without a search
Seems like, if you search for an artist, then with a high probability, you will also find some or all of their albums as well. When this happens, `spt` will asynchronously dispatch an event for populating its knowledge of the user's saved albums.

This works pretty well most of the time, but has two main issues:

1) If you navigate to an artist's page outside of a search, e.g., through your saved artists, your saved albums won't be recognized.
2) If by some quirk of Spotify's search, one or more of the artist's albums is not fouund, and you eventually load the page for that artist, the correct albums won't be recognized as saved.

Here's a screenshot of an artist I follow, that I found by searching:
image
Note that the saved albums are starred.
**NOTE** I'm running a fork of `spt` where I have changed the glyph used for saved albums, because the heart wasn't rendering with the font I use. I'm considering filing another issue to make the symbol configurable for those who have non-unicode supporting fonts/environments like me, or may dislike the icon, etc, but this is not relevant.

Here's a screenshot of the _same_ artist that I navigated to from my saved artists:
image

**NOTE** I feel like it would make sense that the code would be analogous for saved tracks as well, but the current code only dispatches an event for caching the saved albums _only_. I'm assuming this omission was intentional, but it is a little odd so this issue may be applicable to tracks too.

Utagai:
This closes #611 which I just filed.

Check out the issue for some screenshots. This PR effectively makes the second screenshot in that issue description correct, in that it recognizes the saved albums on that artist.

As mentioned in that issue too, we seemingly _do not_ update the saved tracks set when searching, so I omitted adding that into this patch, but it's easy enough to extend this patch to apply to tracks and make things consistent.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -785,6 +785,13 @@ impl<'a> Network<'a> {
 
     if let Ok((albums, top_tracks, related_artist)) = try_join!(albums, top_tracks, related_artist)
     {
+      let album_ids = albums
+        .items
+        .iter()
+        .filter_map(|item| item.id.to_owned())
+        .collect();
+      self.current_user_saved_albums_contains(album_ids).await;
+
       let mut app = self.app.lock().await;
       app.artist = Some(Artist {
         artist_name,
@@ -979,10 +986,10 @@ impl<'a> Network<'a> {
   }
 
   async fn user_artist_check_follow(&mut self, artist_ids: Vec<String>) {
-    if let Ok(are_follwed) = self.spotify.user_artist_check_follow(&artist_ids).await {
+    if let Ok(are_followed) = self.spotify.user_artist_check_follow(&artist_ids).await {
       let mut app = self.app.lock().await;
       artist_ids.iter().enumerate().for_each(|(i, id)| {
-        if are_follwed[i] {
+        if are_followed[i] {
           app.followed_artist_ids_set.insert(id.to_owned());
         } else {
           app.followed_artist_ids_set.remove(id);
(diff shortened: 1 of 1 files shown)


src/network.rs
```diff
@@ -776,6 +776,15 @@ impl<'a> Network<'a> {
     if let Ok((albums, top_tracks, related_artist)) = try_join!(albums, top_tracks, related_artist)
     {
       let mut app = self.app.lock().await;
+
+      app.dispatch(IoEvent::CurrentUserSavedAlbumsContains(
```
Utagai, line 780: @Rigellute I'm curious on your opinion here. If you look at the initial commit for this patch, you'll see I took the more direct approach of calling `current_user_saved_albums_contains()` here to be more direct.

I later changed it to dispatch a `IoEvent::CurrentUserSavedAlbumsContains` because this seemed more robust and consistent, and also avoids the brittleness of the prior approach where we had to carefully call `current_user_saved_albums_contains()`` prior to when we take the lock here.

Utagai:
@Rigellute Seems like I'm failing the `clippy` tests, but for things unrelated to this patch. Is this some new issue that `clippy` is now catching? Seems like we could be using a more concise macro for the key mapping business, not entirely sure.

Maybe this warrants a separate PR?

src/network.rs
sputnick1124: The dispatch won't block whereas we'd have to `await` the `current_user_saved_albums_contains` call which could take some, so dispatching the event is the right move to make here to prevent any blocking in this function due to network IO latency.

Utagai: Thanks, I am new to rust so I didn't feel confident.

Rigellute (approved):
Good work @Utagai, this is a nice improvement. 

I think clippy is failing now due to the recent Rust update. I'll fix, and let you know when you can rebase.

Rigellute:
You should be able to rebase now and fix the failing CI.

Utagai:
@Rigellute Done.

Rigellute:
@all-contributors please add @Utagai for code

PR #614, opened 2020-10-11
Fix new clippy warnings

Rigellute:
The latest rust upgrade brought some new warnings from clippy. This PR fixes them.


Rigellute:
Unfortunately, clippy fails with a panic on CI https://github.com/rust-lang/rust-clippy/issues/6147

So I've needed to downgrade just clippy to run on the 1.47.0 toolchain.

PR #623, opened 2020-10-16
Support searching by Spotify share URLs and URIs analogously to the desktop client

### LINKED ISSUE #41: Open playlists by url?
Would love to know if I am listening to the correct playlist when a friend shares one, such as https://open.spotify.com/playlist/0LbHaoYiD4HCZrrHsf5JK5?si=LlvBb9ySSDmNqzuH4qhRiw

Utagai:
This closes #41.

To be more precise, this actually adds significantly more functionality than what is requested in #41. In particular, this PR adds support for opening:

- Tracks (will open its containing album and move the selection to the searched track)
- Albums
- Artists
- Playlists
- Shows/Podcasts

Via either their Spotify share URL, e.g.:

```
https://open.spotify.com/album/6QdCohkHKNTVoaSx1ZzitH?si=vJQhe4H5RwiGcDf4xr9DuQ
```

Or their Spotify URI:
```
spotify:album:6QdCohkHKNTVoaSx1ZzitH
```

On the official desktop client, searching with either of these leads you to the Metallica album, [Metallica](https://open.spotify.com/album/6QdCohkHKNTVoaSx1ZzitH).

If the given input string fails to match either URI format, it will default to the original raw phrase search.

Here are some representative screenshots:
image
Press enter...
image
**Note** The highlighting/focusing of the track is automatically done.

There were some decisions I made in this PR that may be worth discussing. For those, I've let a comment.

### DIFF
--- a/src/handlers/input.rs
+++ b/src/handlers/input.rs
@@ -164,10 +164,10 @@ fn attempt_process_uri(app: &mut App, input: &str, base: &str, sep: &str) -> boo
     return true;
   }
 
-  let (podcast_id, matched) = spotify_resource_id(trimmed_uri, "podcast");
+  let (show_id, matched) = spotify_resource_id(trimmed_uri, "show");
   if matched {
-    eprintln!("matched podcast, open for id: {}!", podcast_id);
-    // TODO(may): Dispatch some kind of GetPodcast event, analogous to the ones above?
+    eprintln!("matched show, open for id: {}!", show_id);
+    app.dispatch(IoEvent::GetShowEpisodes(show_id));
     return true;
   }
 
--- a/src/handlers/input.rs
+++ b/src/handlers/input.rs
@@ -117,10 +117,6 @@ fn process_input(app: &mut App, input: String) {
 fn spotify_resource_id(base: &str, uri: &str, sep: &str, resource_type: &str) -> (String, bool) {
   let uri_prefix = format!("{}{}{}", base, resource_type, sep);
   let id_string = uri.trim_start_matches(&uri_prefix);
-  eprintln!(
-    "uri prefix found: {}, for uri: {}, id_string: {}",
-    uri_prefix, uri, id_string
-  );
   // If the lengths aren't equal, we must have found a match.
   (id_string.to_string(), id_string.len() != uri.len())
 }
@@ -142,21 +138,18 @@ fn attempt_process_uri(app: &mut App, input: &str, base: &str, sep: &str) -> boo
 
   let (track_id, matched) = spotify_resource_id(base, input, sep, "track");
   if matched {
-    eprintln!("matched track, open for id: {}!", track_id);
     app.dispatch(IoEvent::GetAlbumForTrack(track_id));
     return true;
   }
 
   let (playlist_id, matched) = spotify_resource_id(base, input, sep, "playlist");
   if matched {
-    eprintln!("matched playlist, open for id: {}!", playlist_id);
     app.dispatch(IoEvent::GetPlaylistTracks(playlist_id, 0));
     return true;
   }
 
   let (show_id, matched) = spotify_resource_id(base, input, sep, "show");
   if matched {
-    eprintln!("matched show, open for id: {}!", show_id);
     app.dispatch(IoEvent::GetShowEpisodes(show_id));
     return true;
   }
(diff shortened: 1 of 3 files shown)


src/handlers/input.rs
```diff
@@ -121,6 +94,74 @@ pub fn handler(key: Key, app: &mut App) {
...
+
+  // Default fallback behavior: treat the input as a raw search phrase.
+  app.dispatch(IoEvent::GetSearchResults(input, app.get_user_country()));
+  app.push_navigation_stack(RouteId::Search, ActiveBlock::SearchResultBlock);
+}
+
+fn spotify_resource_id(base: &str, uri: &str, sep: &str, resource_type: &str) -> (String, bool) {
+  let uri_prefix = format!("{}{}{}", base, resource_type, sep);
+  let id_string_with_query_params = uri.trim_start_matches(&uri_prefix);
+  let query_idx = id_string_with_query_params
```
Utagai, line 120: I've made a source-code comment on this, but at some point we may be re-inventing the wheel with regards to parsing URIs.

At the moment, my strong belief is that this should actually be pretty robust and handle all cases of Spotify share URLs, but in the case that it doesn't we should introduce the `url` crate to parse out share-URLs as a special case.

src/network.rs
```diff
@@ -377,9 +381,7 @@ impl<'a> Network<'a> {
 
       let mut app = self.app.lock().await;
       app.playlist_tracks = Some(playlist_tracks);
-      if app.get_current_route().id != RouteId::TrackTable {
-        app.push_navigation_stack(RouteId::TrackTable, ActiveBlock::TrackTable);
-      };
+      app.push_navigation_stack(RouteId::TrackTable, ActiveBlock::TrackTable);
```
Utagai, line 384: I believe this was a bug that I noticed while working on this patch, and is technically not related to the core issue.

This conditional push to the navigation stack made it so that:
- On the first playlist navigation, the focus would correctly and as expected, shift to the playlist window/frame.
- If the user were to immediately select another playlist to change focus to, the playlist would render, but the focus would not shift to the playlist.

Technically, it seems like this would happen whenever the user selected a playlist while on a track table.

src/network.rs
```diff
@@ -1246,6 +1248,39 @@ impl<'a> Network<'a> {
...
+          let selected_album = SelectedFullAlbum {
+            album,
+            // Overflow should be essentially impossible here, so we prefer the cleaner 'as'.
+            selected_index: zero_indexed_track_number as usize,
+          };
+
+          let mut app = self.app.lock().await;
+
+          app.selected_album_full = Some(selected_album.clone());
+          app.saved_album_tracks_index = selected_album.selected_index;
```
Utagai, line 1273: This was something that was necessary to get the auto-focus to move to the selected track.

That said, I think there may be some improvement we can make regarding this global variable. From what I can tell, it seems like for `Simplified` `AlbumTableContext`'s, we intuitively use `app.selected_album_simplified.selected_index`, whereas for the `Full` case, we instead use this special `app.saved_album_tracks_index`. This was quite confusing to me and led me to wondering why my change without this L1273 was not working. I think the fact that I had this moment of confusion is probably a sign of a code smell.

I gave a shot myself at removing this and relying on `app.selected_album_full`, and it actually seems to have worked! However, since it was a bit larger a fix/improvement, I did not want to pollute the diff here.

In closing, @Rigellute, please let me know if you think the work of getting rid of this global index state is worth it, and I'll take a crack at it.

Utagai (commented):
Some comments that I think are worth calling out on this patch.

src/handlers/input.rs
```diff
@@ -357,4 +398,109 @@ mod tests {
...
+    }
+
+    #[test]
+    fn mismatched_resource_types_do_not_match() {
+      let playlist_url =
+        "https://open.spotify.com/playlist/1cJ6lPBYj2fscs0kqBHsVV?si=OdwuJsbsSeuUAOadehng3A";
+      let (_, matched) = spotify_resource_id(URL_BASE, playlist_url, "/", "album");
+      assert_eq!(matched, false);
+    }
+  }
```
Rigellute, line 505: Nice tests

src/handlers/input.rs
Rigellute: We can add the `url` crate if it makes things easier, but this this looks fine to me.

src/network.rs
Rigellute: I can't remember why we had the condition there in the first place. Perhaps it was indeed a mistake.

src/network.rs
Rigellute: Yeah, tracking indexes of the UI is pretty messy at the moment. I've long wanted to go back an look for a better solution. I think the `tui` crate might have newer ways of dealing with this, but I need to check it out.

Rigellute (approved):
👏 good work @Utagai. This will be a very welcome enhancement. 

I also appreciate the clarity of context you've provided with comments in code/github.

Rigellute:
If you can please rebase, I'll merge!

Utagai:
@Rigellute Done.

Rigellute:
Thanks again @Utagai

PR #635, opened 2020-10-26
WIP: Improve results paging

sputnick1124:
I've been doing quite bit of work on trying to unify the paging interface and behavior the past couple of weekends. This is still very much a work in progress, but I would love to get some feedback from other developers if you think the current path might make things better. 

Three are a few things that are a little more complicated than I would like them to be, but the functionality is shaping up to be what I envisioned. 

~~Currently, I have only implemented paging for for the podcast episode view, but now that it is working, I should be able to implement the same for all of the library items as well pretty easily. Then I can move on to playlists and albums.~~

~~It's a bit late here right now, so I will update this description with a to-do list in the coming days to track progress. Also, I'm aware of the falling CI checks, those will be fixed before this is ready for complete review.~~

# TODO:
### TrackTable (delegate track context management to the type system via type aliases?)
- [ ] MyPlaylists
- [ ] AlbumSearch
- [ ] PlaylistSearch                                                            
- [x] SavedTracks
- [ ] RecommendedTracks
- [ ] MadeForYou
### Search results
- [ ] Albums
- [ ] Artists
- [ ] Tracks
- [ ] Playlists
- [ ] Podcasts
### Library Collections
- [x] Saved artists
- [x] Saved albums
- [x] Saved tracks
- [x] Made for you
- [x] Podcasts
### Misc
- [ ] User playlists
## Other
- [ ] Library podcasts view (needs to happen before we can page results)

# Call to action
Please review the changes at any point. This PR represents a massive change to how we are handling almost every list of items in the app; I'd like to get it right so we aren't plagued with issues in the future. 

Also, let me know if there is any thing that is currently not being paged that is eligible for paging that is not captured in the to-do list above.

If anyone would like to help out, much of the work could be done in parallel from here, I think. Let me know and we can coordinate. We could either work on my fork directly, or we could target a feature branch here.


src/handlers/artists.rs
```diff
@@ -7,52 +7,21 @@ use crate::{
...
-        let next_index = common_key_events::on_middle_press_handler(&artists.items);
-        app.artists_list_index = next_index;
-      }
-    }
-    k if common_key_events::low_event(k) => {
-      if let Some(artists) = &mut app.library.saved_artists.get_results(None) {
-        let next_index = common_key_events::on_low_press_handler(&artists.items);
-        app.artists_list_index = next_index;
-      }
+    k if common_key_events::is_list_navigation_key_event(k, app) => {
```
Rigellute, line 10: `is_list_navigation_key_event` looks like a great abstraction

src/paging.rs
```diff
@@ -0,0 +1,174 @@
...
+      }
+      k if common_key_events::list_end_event(k) => self.items.len() - 1,
+      k if common_key_events::list_begin_event(k) => 0,
+      k if common_key_events::is_list_navigation_key_event(k, app) => {
+        unimplemented!("List navigation event {:?} needs an implementation", k)
+      }
+      _ => unreachable!("This function cannot handle non-navigation key events!"),
+    }
+  }
+}
```
Rigellute, line 174: This is looking good

src/app.rs
```diff
@@ -665,123 +662,12 @@ impl App {
...
-      }
-    }
-  }
-
-  pub fn get_current_user_saved_albums_previous(&mut self) {
-    if self.library.saved_albums.index > 0 {
-      self.library.saved_albums.index -= 1;
-    }
-  }
-
```
Rigellute, line 784: Will be nice to remove all this from `app`

Rigellute (commented):
This is looking promising @sputnick1124!

sputnick1124 (commented):
I think I may be noticing a degradation in performance using this fork. I'd like to hear if others are seeing the same thing.

sputnick1124:
Just a quick update for anyone that may be watching this: I have been super busy as of late and haven't been able to give this much time. I have run into a couple of design issues in implementing paged playlist tracks; I think I have an idea of an elegant solution, but haven't started implementation yet, so we'll see. Once playlist and album tracks are done, I think there won't really be any more surprises and I can zip through the rest of the checklist items.

Issue #533, opened 2020-07-22
Help menu flickers when viewing in vertical orientation


isaacdchan:
Currently have spt on a 1080 x 1920 screen. Problem disappears with small enough font size

Understand vertical orientation usage is probably the minority so no rush.

Screen flickers between these two screens:
![image](https://user-images.githubusercontent.com/33741008/88128760-77dbaf00-cb8b-11ea-9a83-eb08580137e4.png)
![image](https://user-images.githubusercontent.com/33741008/88128822-95107d80-cb8b-11ea-98bb-35256171a7cd.png)
With smaller font:
![image](https://user-images.githubusercontent.com/33741008/88128925-bcffe100-cb8b-11ea-97ce-9e2c3517ae51.png)

Thanks again for such a cool project!

Rigellute:
I also experience this - haven't found a fix yet. 

Thanks for opening the issue 👍

Nil-Cipher:
I'm also having this issue. 
![glitchy](https://user-images.githubusercontent.com/20496131/88466458-787d8980-ce9a-11ea-9a49-9acb594868fd.png)
![normal](https://user-images.githubusercontent.com/20496131/88466471-93e89480-ce9a-11ea-9568-091dc7e77bc1.png)

dp304:
Hi!

I'm also having this with a "standard" 80x24 terminal (my default).

The problem seems to be that the Cassowary solver (which is used by `tui` to calculate the column widths in a table) has some non-determinism if the constraints cannot be satisfied; which is the case when the terminal is too narrow to hold all the columns. 
In its [documentation](https://docs.rs/cassowary/0.3.0/cassowary/#a-simple-example), there is more explanation to this, as well as a suggested solution to a related problem.

Perhaps a similar fix could be applied to `tui`: in [`Table::render()`](https://github.com/fdehau/tui-rs/blob/v0.12.0/src/widgets/table.rs#L244-L268), one could add extra (stronger) constraints on the width ratio of neighbouring columns, but I'm not sure if this would actually work. Possibly related issues in `tui`: fdehau/tui-rs#161, fdehau/tui-rs#314

Until this problem is somehow handled in `tui`, I suggest a workaround (PR coming up), where the rows of the table are formatted "manually" into one column.

Another thing I've tried is playing around with the width constraints. For example, one could use `Constraint::Percentage`s instead of `Constraint::Length`s, but there are some problems with this:
* If the terminal is _very_ narrow, it still flickers.
[...]

PR #638, opened 2020-10-28
Issue #533

dp304:


Rigellute (approved):
Interesting workaround. Thanks for figuring this out @dp304 👍

Issue #523, opened 2020-07-10
Rapidly seeking through song causes song to get "stuck"


dvdmuckle:
If I try to rapidly seek through a song using `<` or `>` numerous times in succession, the song will get stuck at whenever I started seeking and will constantly start seeking to that point instead of seeking back or ahead of that point.

Video: https://asciinema.org/a/o7c0aue0g2np34HCReOQswjwr

I think what's going on is there's a client side calculation of where to seek to (current song position + fifteen seconds), however because the API call takes some time to go through, queuing up multiple seeks doesn't work, as the current song position is retrieved from the API as opposed to some local variable.

Rigellute:
Thank you for the asciinema!

I think a fix for this would be to "debounce" the seek, sum the total, and then make one api request with the final seek

dp304:
See PR #640
There is a little latency before the actual seeking (up to 5s), but on the other hand, it is very light on API calls.
The visual feedback is quasi-instant though.

dvdmuckle:
That PR is exactly the solution to solve this issue, awesome! Once that's merged in this can be closed.

PR #640, opened 2020-10-29
Optimize seek (#523)

dp304:
Instead of immediately sending the API request when '<' or '>' is
pressed, the offsets are accummulated. The seek request is sent
periodically along with the playback update request.

Also, the progress bar in the UI shows the desired seek position even
before the actual seek is performed.


Rigellute (approved):
Tried this out, and looks good! Thanks again @dp304 👍

Rigellute:
@all-contributors please add @dp304 for code

Issue #429, opened 2020-05-03
Provide a cli arg based interface


clearnote01:
I want to trigger commands that goes through spotify-tui but doesn't open the UI. For starters support for search/play/pause through command line would be really good to have.

spt -s --search "a man of wire"  // returns top n results
spt -p --play "song name" // based on closest match, start playing this song
spt -t --toggle // toggle play pause

Rigellute:
Could be an interesting addition 👍 , will require some architectural changes but is doable

dvdmuckle:
It would also be great to have something like `spt --status` to show the currently playing song. 

As an aside, there are several tools that allow a CLI arg based interface for Spotify, but they're very fragmented and aren't all mutliplatform, and generally control Spotify in different ways (DBUS/MPRIS for Linux or oascript for Mac) as opposed to interfacing with the Spotify API.

Dragonsong3k:
> I want to trigger commands that goes through spotify-tui but doesn't open the UI. For starters support for search/play/pause through command line would be really good to have.
> 
> spt -s --search "a man of wire" // returns top n results
> spt -p --play "song name" // based on closest match, start playing this song
> spt -t --toggle // toggle play pause

Launching a playlist in shuffle mode would be great!
spt -p "some playlist name" -s ( --shuffle)

dann-merlin:
I think that sounds more like a separate program. Something like this [script](https://gist.github.com/wandernauta/6800547) referenced by spotifyd's README, but more powerful.

OrangeFran:
I picked this up yesterday. Expect a demo this evening :)

OrangeFran:
Never used imgur, I hope it works: [here](https://imgur.com/a/eE7ft1g).

OrangeFran:
I actually did not have to change anything internally. Currently I'm not really happy with the cli.rs file, it seems so unordered. I have to look into that. If you want to try it out, you can use the cli branch on my fork.

OrangeFran:
I did some huge changes to the cli:

There are three main subcommands: `play` to play a song, `playback` to interact with the playback and `query` to list and search categories. One of my favourite things is the fact that you can format almost any output with a simple string where `%a` -> artist, `%b` -> album, `%t` -> track, `%h` -> show, `%f` -> flags (like, shuffle, repeat), `%p` -> playlist, `%d` -> current device and `%s` -> playing status. I really hope it is intuitive enough.

spenat28:
[...]

PR #645, opened 2020-11-01
Cli interface

OrangeFran:
Hi there.

First of all, I know this is a lot of code. So sorry. I tried write the code as clean and as easy to understand as possible. Let me know if I can do anything to make this whole process easier for you. I actually did not have to make any architectural changes to the code base. I just use `network.handle_network_event(...).await` directly to query from the server. 

This would implement a basic command line interface to control spotify. #429 

Most of these arguments are self-explanatory. I tried my best with clap Args and ArgGroups.
What I like the most is the `format` option. It replaces format specifiers with their value or 'None'.
Possible specifiers are currently:
```
%a  -> artist
%b  -> album
%t  -> track
%p  -> playlist
%h  -> show
%u  -> uri
%s  -> status (playing or paused) (emoji)
%f  -> flags (like, repeat, shuffle) (emojis)
%v  -> volume
%d  -> current device
``` 

**What needs to be done?**

- [x] Add voume control options (done)
- [x] Fix `--repeat` bug (does not work with RepeatState::Track) (spotifyd bug)
- [x] Set device_id in config if wrong to first device avaible
- [x] Implement anyhow::Errors (done where necessary)
- [ ] Number of search results (maybe a `--offset` flag?)
- [x] Shell completion
- [x] Configurable icons
- [ ] Made for you

* Feature requests?


OrangeFran:
Wait, I think I've messed up something somehow. I have no idea what. There shouldn't be so many changes.

Rigellute:
Sorry for the delay @OrangeFran - I will try to review this as soon as I can

OrangeFran:
No rush 👍

OrangeFran:
If you think this is out of scope (of this project), that's fair. I could create my own repo.

Rigellute (commented):
This is extremely impressive work @OrangeFran. 

I've looked through the code and all looks clean. Only slight concern is size of the feature, but I don't think there's a way around that.

This is a feature that has been requested multiple times and I think it is within scope of the project.

I will test this out and get back to you again. Sorry it is taking so long, I am snowed under at the moment!

src/cli.rs
```diff
@@ -0,0 +1,804 @@
...
+  },
+  senum::RepeatState,
+};
+
+const LIKED_EMOJI: &str = " ";
+const SHUFFLE_EMOJI: &str = "咽";
+const REPEAT_T_EMOJI: &str = "綾";
+const REPEAT_C_EMOJI: &str = "凌";
+const PLAYING_EMOJI: &str = "契";
+const PAUSED_EMOJI: &str = " ";
```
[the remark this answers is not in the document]
OrangeFran: Would be cool. I saw that somebody already made the liked icon and probably more, so I could use that and add some config vars for the new icons as well.

OrangeFran:
How do you stand on the search thing? Do you think we should keep it in or take it out?
It seems way easier to just use the ui for that and I don't see any real use cases, do you?

jaor:
> How do you stand on the search thing? Do you think we should keep it
> in or take it out?  It seems way easier to just use the ui for that
> and I don't see any real use cases, do you?

a thing i'll use the cli interface for is providing some kind of emacs
front-end to spotifyd/spotify-tui.  being able to perform searches would
be very useful in that scenario.

but then i'm not a contributor, so you shouldn't pay much attention to
my needs :)

thanks for your work, btw!

OrangeFran:
The problem is not that you're not a contributor, the problem is that you use emacs :)
Anyway, the feature is already there and if somebody uses it, I see no reason to take it out.
I'm also adding some more options for searching, like specifying a limit of results, and probably an offset for more fine-grain control.

src/main.rs
```diff
@@ -140,6 +162,187 @@ async fn main() -> Result<()> {
...
+                                    .long("shows")
+                                    .help("Search the category 'shows'"))
+                               .group(ArgGroup::with_name("searchable")
+                                    .args(&["playlists", "tracks", "albums", "artists", "shows"])
+                                    .multiple(false))
+                               // Actions
+                               .group(ArgGroup::with_name("actions")
+                                    .args(&["list", "search"])
+                                    .multiple(false)
+                                    .required(true)))
```
Rigellute, line 345: Would you be able to move this clap construct into in the cli module please? Just because it's so large 😄.

Could create a `cli` dir with `cli.rs` and `clap.rs` or something like that. We could use this module later to further breakdown the `cli.rs` file.

Rigellute: Also, I've been meaning to try out [structopt](https://github.com/TeXitoi/structopt), which might simplify this.

But don't worry about that for this PR.

Rigellute (commented):
Have just tried it out and works really well! Just one comment about moving the `clap` construct out of `main.rs` and into the cli module.

src/main.rs
OrangeFran: Makes sense, it's pretty big.

OrangeFran:
New: configurable icons and a `--limit` flag to change the query limits.
I changed the default liked icon to something bigger because the cli looked better with it.
If it doesn't match your preference, we could just change it back, or add a seperate option for cli icons.

OrangeFran:
While working on this (getting an url for the current track/album), I found out that the copied album url is actually wrong. It's not `https://open.spotify.com/album/{track.id}` but `https://open.spotify.com/album/{track.album.id}`. This is an easy fix tho.

OrangeFran:
Regarding the size of this feature: It seems that a lot of the functionality is already implemented, but in a slightly different way for the UI so the CLI can't use it. A possible way to reduce the size would be to split up some of the `App` functions into smaller functions with the bare functionality so that the CLI and the UI can use them. I am a bit terrified though of bricking something and it would probably be a lot more tedious to review. I also don't know where these functions should go.

Let me know what you prefer.

Rigellute (approved):
This is an outstanding contribution @OrangeFran 👏. 

I agree that we could share lots of code between UI and CLI, and perhaps we could tackle that after merging this. 

I will merge now, thanks again 🙏.

PR #650, opened 2020-11-05
Implement Podcasts Library page

LeonGr:
Implemented:
- Podcasts page shows list of user's saved podcasts (columns: name, publisher(s))
- Pressing "Enter" on podcast goes to episode table
- Pressing "D" unsaves selected podcast

Functions related to podcasts that are not yet implemented:

- ~~Showing heart next to followed podcasts in search results~~
- ~~Ability do follow/unfollow from search results~~
- Ability to follow/unfollow from episode table

Possible additions:
- Add name of podcast to episode table

I'm willing to work on these other functionalities too but I thought it would be good to get feedback for what I have already done. 
And I was wondering if it was okay to add a `selected_show_id` field to `App`, because to implement following/unfollowing from the episode table we need the viewed show's id. Another way would be calling `get_an_episode` from rspotify to get the `FullEpisode` instead of `SimplifiedEpisode` of which the former includes a reference to the show but this might create unnecessary overhead.

EDIT:
Hearts and follow/unfollow now works in podcasts search results.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -11,7 +11,7 @@ use rspotify::{
     page::{CursorBasedPage, Page},
     playing::PlayHistory,
     playlist::{PlaylistTrack, SimplifiedPlaylist},
-    show::{SimplifiedEpisode, SimplifiedShow, Show},
+    show::{Show, SimplifiedEpisode, SimplifiedShow},
     track::{FullTrack, SavedTrack, SimplifiedTrack},
     user::PrivateUser,
     PlayingItem,
--- a/src/app.rs
+++ b/src/app.rs
@@ -11,7 +11,7 @@ use rspotify::{
     page::{CursorBasedPage, Page},
     playing::PlayHistory,
     playlist::{PlaylistTrack, SimplifiedPlaylist},
-    show::{SimplifiedEpisode, SimplifiedShow},
+    show::{SimplifiedEpisode, SimplifiedShow, Show},
     track::{FullTrack, SavedTrack, SimplifiedTrack},
     user::PrivateUser,
     PlayingItem,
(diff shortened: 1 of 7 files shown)


src/app.rs
```diff
@@ -973,12 +1002,46 @@ impl App {
...
   }
 
-  pub fn user_follow_show(&mut self) {
-    unimplemented!();
+  pub fn user_follow_show(&mut self, block: ActiveBlock) {
+    match block {
+      ActiveBlock::SearchResultBlock => {
+        if let Some(shows) = &self.search_results.shows {
+          if let Some(selected_index) = self.search_results.selected_shows_index {
+            let show_id = shows.items[selected_index].id.to_owned();
```
Rigellute, line 1010: This could use `get` in case the index is somehow out of range?

```suggestion
            if let Some(show_id) = shows.items.get(selected_index).map(|item| item.id) {
               ...
            };
```

src/ui/mod.rs
```diff
@@ -510,7 +511,14 @@ where
       Some(podcasts) => podcasts
         .items
         .iter()
-        .map(|item| format!("{:} - {}", item.name, item.publisher))
+        .map(|item| {
+          let mut show_name = String::new();
+          if app.saved_show_ids_set.contains(&item.id.to_owned()) {
+            show_name.push_str("♥ ");
```
Rigellute, line 517: The "liked" icon is now configured in the user config which can be reached by

```rs
&app.user_config.padded_liked_icon()
```

src/ui/mod.rs
```diff
@@ -1104,36 +1164,36 @@ where
...
+//     .border_style(get_color(highlight_state, app.user_config.theme));
+//
+//   let text = Text::from("Not implemented yet!");
+//
+//   let not_implemented = Paragraph::new(text)
+//     .style(Style::default().fg(app.user_config.theme.text))
+//     .block(display_block)
+//     .wrap(Wrap { trim: true });
+//   f.render_widget(not_implemented, layout_chunk);
+// }
```
Rigellute, line 1196: Did you mean to comment this out? If it's no longer needed, could you delete?

Rigellute (commented):
Thanks for working on this @LeonGr - just left a few comments

Rigellute (approved):
Looks good, thanks for this @LeonGr

PR #659, opened 2020-11-14
Make the liked icon configurable

### LINKED ISSUE #625: Make the liked icon configurable
Currently, the icon `spt` uses to mark liked items, such as a liked song, is the heart character: ♥.

For any variety of reasons, the user may want to change their preferred icon. e.g.:
1. Their font family does not support the glyph.
2. Their terminal does not render it properly.
3. Preference

For me, #1 applies, so I unfortunately have an empty space whenever I see a liked song. I was thinking that we could make the default value stay as the ♥, but allow users to configure a replacement string if they wish.

Utagai:
This closes #625.

Sorry this took longer than it should have. I got distracted.

Anyways, it's a pretty simple change. I only have one reservation about the naming.

Here's a screenshot of it in action:
image

And here's the accompanying config:
```yaml
theme:
  ...

behavior:
  liked_icon: "★"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,8 +223,11 @@ behavior:
   tick_rate_milliseconds: 250
   # Enable text emphasis (typically italic/bold text styling). Disabling this might be important if the terminal config is otherwise restricted and rendering text escapes interferes with the UI.
   enable_text_emphasis: true
-  # controls whether to show a loading indicator in the top right of the UI whenever communicating with Spotify API
+  # Controls whether to show a loading indicator in the top right of the UI whenever communicating with Spotify API
   show_loading_indicator: true
+  # Determines the text icon to display next to "liked" Spotify items, such as
+  # liked songs and albums, or followed artists. Can be any length string.
+  liked_icon: "♥"
 
 keybindings:
   # Key stroke can be used if it only uses two keys:
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -395,7 +395,7 @@ where
             song_name += "▶ "
           }
           if app.liked_song_ids_set.contains(&id) {
-            song_name += "♥ ";
+            song_name += app.user_config.liked_icon();
           }
 
           song_name += &item.name;
@@ -423,7 +423,7 @@ where
         .map(|item| {
           let mut artist = String::new();
           if app.followed_artist_ids_set.contains(&item.id.to_owned()) {
-            artist.push_str("♥ ");
+            artist.push_str(app.user_config.liked_icon());
           }
           artist.push_str(&item.name.to_owned());
           artist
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -223,8 +223,11 @@ behavior:
   tick_rate_milliseconds: 250
   # Enable text emphasis (typically italic/bold text styling). Disabling this might be important if the terminal config is otherwise restricted and rendering text escapes interferes with the UI.
   enable_text_emphasis: true
-  # controls whether to show a loading indicator in the top right of the UI whenever communicating with Spotify API
+  # Controls whether to show a loading indicator in the top right of the UI whenever communicating with Spotify API
   show_loading_indicator: true
+  # Determines the text icon to display next to "liked" Spotify items, such as
+  # liked songs and albums, or followed artists. Can be any length string.
+  liked_icon: "♥"
```
Utagai, line 230: My reservation is here. Technically, in almost all cases this is indeed the "liked icon". However, for artists, one "follows" them, so it kind of breaks there.

I considered some alternative names like "tracked_icon" or similar, but it just didn't feel right. Open to any renaming ideas.

OrangeFran:
Hey. 

I really like this. Thank you. Would you like to **implement this for other icons, like shuffle, repeat and currently playing**. I'm working on a cli and using icons a lot so I'm gonna implement that.

Utagai:
@OrangeFran It's unclear from your message, are you asking if I'd be interested in implementing custom icons for other things besides this, or are you saying _you're_ going to be doing that?

OrangeFran:
Sorry, I meant that I'm gonna change from static icons to the configured ones in my fork of the project after this PR gets merged. I wanted to ask if you would be interested in doing the same for other icons. Hopefully this cleares things up.

Utagai:
I think I get it. Yeah, I was thinking of doing it for the current arrow marker as well. Not aware of other icons though. Granted, I've never tried shuffle. You mention for example, repeat, but there is no repeat text icon. Does that mean you're suggesting we add one? For example, Repeat currently just uses a string, e.g., `"All"`, `"Track"`, etc.

OrangeFran:
Yeah, adding a repeat icon would be cool. But there probably should be a seperate PR for that. I thought they were already there, my bad.

Utagai:
No worries. But yes, there really aren't any other icons besides the one made configurable in this PR and the current marker (arrow). We _could_ perhaps add configurable ones for repeat/shuffle, but I'm unsure of what valid options there could be. For example, a simple search for "repeat" and "loop" over the unicode characters do not reveal many sensible options. Especially considering that repeat has various modes, it becomes even harder to distinguish them.

Did you have particular icons in mind?

OrangeFran:
True, I thought of nerd font icons. They have pretty nice shuffle and repeat icons. But there are no unicode icons, at least I didn't find any. Maybe that's the reason that there aren't any more icons implemented.

Utagai:
Yeah, nerd font, font awesome, etc, could open the doors to something that applies.

In either case, definitely doable. This PR shows it shouldn't be too bad to extend to a few more possibilities. I'm open to doing it if there's interest and @Rigellute is OK with it.

Rigellute (approved):
Great work @Utagai, this looks like a nice addition. Thank you 🙏

PR #668, opened 2020-11-27
Handle invalid Client ID/Secret

aokellermann:
### Changes
* Asks user for Client ID/Secret until a valid input is entered (32 hex digits)

This makes it such that the config file is not written unless valid inputs are entered, so the user will not have to manually edit or delete the config file.


Rigellute (approved):
So sorry for the extremely late review. This looks great! Thank you 👍

Rigellute:
@all-contributors please add @aokellermann for code

PR #673, opened 2020-12-05
cd: extract variables in order to simplify script

MarcoIeni:
Extracting variables like that works for me [here](https://github.com/rust-github/rust-gh-example/blob/01e0aae91f83e1e932f4498a0eddb2905c8fffc3/.github/workflows/cd.yml#L74-L95).


Rigellute (commented):
Nice improvement indeed. Thanks again 👍

PR #703, opened 2021-01-12
Fix Default tick rate

Epacnoss:
In my opinion, fixing the tick rate to be 16ms - roughly 60fps, makes it appear much smoother, without compromising on performance.


Rigellute:
Thanks for looking at this @Epacnoss. 

This is true but in my tests it used up lots of CPU on weaker machines. 

Knowing this however, I set up a cli arg that can configure the tick rate. 

To get 60fps you can pass the arg `--tick-rate 16` on start up

```sh
spt --tick-rate 16
```

If this option wasn't obvious enough, please let me know and we can document it better.

PR #717, opened 2021-01-22
Change `--like` that toggles the state to explicit `--like` and `--dislike` flags

OrangeFran:
Title. After some usage I feel that this makes more sense. Related to this issue: #429, idea from @spenat28.


Rigellute (approved):
Great work once again @OrangeFran. 

And thanks for responding to the issue.

Issue #737, opened 2021-02-12
Search field is now quite small


trames57:
I looked through some of the recent changelog entries and could not find where this was changed, but in v22, my search field/help field is no longer across the top, but very short and to the top left.  Is there a config option that will restore this?

The short search field for me is not what I would like since I search for classical music. It works, but many of the words will become not visible.

I do not have a .config file in my home folder, so I guess I am using whatever the defaults are.

OrangeFran:
I don't think there's an option to disable this feature. If you make the terminal smaller, you will see that the search bar gets wider again. This is due to a change made in winter last year I worked on, because a lot of people don't have long search queries. I opened a pull request to add an option called `enforce_wide_search_bar` to make the search bar wider again. PR: #738

trames57:
Much appreciated! Thanks!

PR #738, opened 2021-02-12
`enforce_wide_search_bar` to make search bar bigger

OrangeFran:
This adds an option with the name `behavior.enforce_wide_search_bar` to make the search bar wider. Fix for #737


Rigellute (approved):
Nice one @OrangeFran, thank you

PR #742, opened 2021-02-22
Replace black and white default colors with reset.

Jesse-Bakker:
Previously, text would be white, which is not visible on bright color
schemes. The color for non-highlighted text now uses the default
foreground color of the users terminal emulator, making it visible in
any theme. The only visual change in dark color schemes is that the
playbar text might go from black to white, depending on the users
terminal emulator (some terminal emulators swap fore- and background
colors on a colored background depending on the background color; this
is IMO the correct behavior anyway).

### DIFF
--- a/src/user_config.rs
+++ b/src/user_config.rs
@@ -53,7 +53,7 @@ impl Default for Theme {
   fn default() -> Self {
     Theme {
       analysis_bar: Color::LightCyan,
-      analysis_bar_text: Color::Black,
+      analysis_bar_text: Color::Reset,
       active: Color::Cyan,
       banner: Color::LightCyan,
       error_border: Color::Red,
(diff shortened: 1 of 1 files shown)


src/user_config.rs
```diff
@@ -53,21 +53,21 @@ impl Default for Theme {
...
+      analysis_bar_text: Color::Reset,
       active: Color::Cyan,
       banner: Color::LightCyan,
       error_border: Color::Red,
       error_text: Color::LightRed,
       hint: Color::Yellow,
       hovered: Color::Magenta,
       inactive: Color::Gray,
-      playbar_background: Color::Black,
+      playbar_background: Color::Reset,
```
Rigellute, line 64: For dark themes, I think this needs to stay as Black otherwise the the text won't be visible when the track progress overlays the text
# playbar_background: Color::Black,
image

# playbar_background: Color::Reset,
image

Rigellute (commented):
Thanks for looking at this. Just one problem I can see. Not sure if there's a general solution for dark/light colour schemes?

src/user_config.rs
Jesse-Bakker: Yeah, I think that can stay black for both dark and light themes (as cyan will probably be pretty close to cyan in most themes and if not, the user can easily fix it themselves with a theme config setting).

Rigellute (approved):
Good work @Jesse-Bakker, thank you

Rigellute:
@all-contributors please add @Jesse-Bakker for code

Issue #762, opened 2021-03-08
Keep receiving API error 404 


FranciscoVasconcelos:
I have a premium account and spotifyd is running in the background and everything works just fine. But this message sistematically appears with no aparent reason. I just ESC and go back to the main page and everything else keeps working (I can still control the music and it keeps playing). This is something that's mostly annoying but it doesn't take away any functionality.
![2021-03-08-080756_954x226_scrot](https://user-images.githubusercontent.com/37560515/110293673-c76dce80-7fe6-11eb-8ae5-06a43e26fd27.png)

sheepwall:
I have the exact same problem.

badele:
> I have the exact same problem.
> edit: release version 0.23.0

Same issue with same version

Rigellute:
Hmm... perhaps the spotify api has changed 🤔

avijit1258:
Same issue here.

lanej:
Same, but sometimes a 404 and sometimes a 503, and it never seems to effect playback.
![image](https://user-images.githubusercontent.com/79638/112513382-851bef80-8d51-11eb-882b-ce6bbb670450.png)
* I have a premium account
* Device is active and has no spaces in the name

lanej:
I'm testing out #772, which is a hack to dump errors to the log instead of consuming the entire ui.

ericlovesmath:
I'm only getting 503 errors, but it only seems to happen in version 0.23.0

ECon87:
Same error here

yaflow-droid:
![image](https://user-images.githubusercontent.com/63092729/113626364-fc0e8d80-9659-11eb-833d-657e63301fbe.png)

I'm getting the same issue, but with a different error message.
It only appears when using spotifyd as a player, and I didn't get the error with the spotify GUI client

arnarg:
At the very least it would be very nice if these error screens weren't stacked and require me to press Esc 50 times to get to the player again.

dimas-cyriaco:
I'm having this issue too. On Mac with `spotify-tui` installed via homebrew.

![Screen Shot 2021-08-05 at 15 53 51](https://user-images.githubusercontent.com/65256/128405412-6e9970ed-526f-449e-a51c-6e2702ced9f4.png)

This happens no matter what device I choose, either `spotifyd`, the web client or the desktop client.

fosBHY:
[...]

PR #772, opened 2021-03-25
feat: dump errors to log, rather than consume ui

lanej:
This is meant to help with #762.  I'll add tests and make it less of a hack if proven functional.

### DIFF
--- a/src/user_config.rs
+++ b/src/user_config.rs
@@ -457,6 +457,10 @@ impl UserConfig {
       self.behavior.repeat_context_icon = repeat_context_icon;
     }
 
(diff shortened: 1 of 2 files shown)


lanej:
This is meant to help with #762.  I'll add tests and make it less of a hack if proven functional.

src/user_config.rs
```diff
@@ -235,6 +236,7 @@ pub struct BehaviorConfig {
   pub repeat_context_icon: String,
   pub playing_icon: String,
   pub paused_icon: String,
+  pub error_log: Option<String>,
```
lanej, line 239: TODO: just create a logger around the described file.

Issue #576, opened 2020-09-13
Closing web client does not stop the play timer


meerasndr:
I'm not sure if the web app qualifies as an official Spotify client, but here goes:
When I close the Spotify tab on my browser, the play countdown keeps running in the terminal till the end of the track length, even though the audio itself stops. Tried this on both Firefox and Brave with the same result.
Is this a known issue? 
PS- I'm able to start and pause from the Terminal, no problems. It displays the client correctly( Web player(Firefox)).

sputnick1124:
Yep. It looks like we don't properly handle the possibility of the playback context disappearing: https://github.com/Rigellute/spotify-tui/blob/0e8f3723ad3b91725d04b460ae6b511ac6ffac1c/src/network.rs#L310-L325

This should be as simple as adding an 
```
} else {
    app.current_playback_context = None;
}
```

Might be a great first issue for someone to tackle. I would be happy to help out if needed.

meerasndr:
Thanks. I can take a look at it!

Rigellute:
Nice!

PR #773, opened 2021-03-28
End playbar continuing when client is closed

Hwatwasthat:
Relevant to [#576](https://github.com/Rigellute/spotify-tui/issues/576). Have just updated as suggested in that, setting the playback context to none. Has dragged some commits in from an earlier pull request that was merged, not sure why.


Hwatwasthat:
I see that clippy is aggressive as last time.

PR #775, opened 2021-03-29
feat: Allow wayland builds without xcb dependency

xanderio:
The arbout carte is used to interact with the system clipboard. This
however imposes a dependency on xcb on the crate for Linux/Unix builds.

On a pure Wayland system this is undesirable as interaction with the
system clipboard via xcb in only possible if the user runs a xwayland
server. This cause spotify-cli to require a dependency to a x11 c lib
without providing any functionally to the user.

This change intoduces to new cargo features to spotify-cli: `x11` and
`wayland`.

The `x11` feature is in the default feature and will preserve the
current application behavior.

The `wayland` feature will disable the dependency onto arboard and use
`#[cfg(feature = "wayland")]` attributes to disable code using this
crate, as well as any reference from the help menu to any copy
functionally.

I currently plan to investigate how to bring this behavior back to the wayland platform.

Looking forward to your code review. :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,11 +23,16 @@ dirs = "3.0.1"
 clap = "2.33.3"
 unicode-width = "0.1.8"
 backtrace = "0.3.56"
-arboard = "1.1.0"
+arboard = { version =  "1.1.0", optional = true }
 crossterm = "0.19"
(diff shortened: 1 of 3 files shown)


xanderio:
I'm unsure what to do about the cipply failure, as this seams to be a general repository issue and nothing particular with this PR

PR #776, opened 2021-03-31
feat: allow clipboard access under wayland

xanderio:
The arbout crate is used to interact with the system clipboard. This
however imposes a dependency on xcb on the crate for Linux/Unix builds.

On a pure Wayland system this is undesirable as interaction with the
system clipboard via xcb in only possible if the user runs a xwayland
server. This cause spotify-cli to require a dependency to a x11 c lib
without providing any functionally to the user.

This change intoduces to new cargo features to spotify-cli: `x11` and
`wayland`.

The `x11` feature is the 'old' default feature and will preserve the
current application behavior.

The `wayland` feature will disable the dependency onto arboard and use
`wl-clipboard-rs` to interact with the wayland clipboard via the
`data_control` protocol. This protocol is part of wlroots.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,7 +27,7 @@ version = "0.7.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "043164d8ba5c4c3035fec9bbee8647c0261d788f3474306f93bb65901cae0e86"
 dependencies = [
- "memchr",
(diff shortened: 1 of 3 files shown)


xanderio:
This PR cloud be replace with an update of `arboard` if this PR gets merged. 
https://github.com/ArturKovacs/arboard/pull/12

Issue #821, opened 2021-05-19
Pressing `d` twice requires `q` twice to exit


RagnarGrootKoerkamp:
The device selector popup can be opened multiple times by pressing `d` more than once.
To get back to the main menu, each of these needs to be closed separately by pressing `q` the same number of times.

It would be nicer if pressing `q` only once always goes back to the main menu.

E.g, actual behavior:
- press `d`
- press `d`
- press `q`
- You're seeing the device selector

expected behavior:
- press `d`
- press `d`
- press `q`
- You're seeing the main menu

PR #826, opened 2021-06-01
Pressing `d` twice requires `q` twice to exit

cobbinma:
Hi guys 👋,

First attempt at a PR here and attempted to resolve [issue](https://github.com/Rigellute/spotify-tui/issues/821)

Did this by `dedup`ing any routes on the navigation stack with matching ids.

---

The device selector popup can be opened multiple times by pressing `d` more than once.
To get back to the main menu, each of these needs to be closed separately by pressing `q` the same number of times.

It would be nicer if pressing `q` only once always goes back to the main menu.

E.g, actual behavior:

press `d`
press `d`
press `q`
You're seeing the device selector
expected behavior:

press `d`
press `d`
press `q`
You're seeing the main menu

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -626,19 +626,18 @@ impl App {
   // The navigation_stack actually only controls the large block to the right of `library` and
   // `playlists`
   pub fn push_navigation_stack(&mut self, next_route_id: RouteId, next_active_block: ActiveBlock) {
-    if self
+    if !self
       .navigation_stack
       .last()
       .map(|last_route_id| last_route_id.id == next_route_id)
       .unwrap_or(false)
     {
-      return;
+      self.navigation_stack.push(Route {
+        id: next_route_id,
+        active_block: next_active_block,
+        hovered_block: next_active_block,
+      });
     }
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -634,6 +634,8 @@ impl App {
   }
 
   pub fn pop_navigation_stack(&mut self) -> Option<Route> {
+    self.navigation_stack.dedup_by(|a, b| a.id == b.id);
```
RagnarGrootKoerkamp, line 637: I think it would be just a bit nicer to do this in `push_navigation_stack`, assuming that that is the only place where the stack is stack is grown. In that case, it could possibly be just an `if` statement checking whether the current id is the same as the id at the top of the stack.

cobbinma: thanks for looking at this 👍 moved to `push_navigation_stack` and changed to if statement.

Rigellute (approved):
Nice, looks good!

Thank you 👏

Rigellute:
@all-contributors please add @cobbinma for code

PR #844, opened 2021-07-29
Add option to set window title to "spt - Spotify TUI" on startup

diegov:
This adds an option to set the window title to "Spotify TUI" on startup. 

This is very handy when switching windows via something like dmenu, since it makes the terminal window where spotify-tui is running easy to find. 

There are a few choices that I wasn't sure about though:

- I've added the configuration option under "behavior", is there a better place for it? 
- I've set it to enabled by default, since it's a very well supported xterm feature so it shouldn't cause problems for most people. Should it default to false to maintain existing behavior on upgrade?
- Since the binary is called spt, i've set the title to "spt - Spotify TUI", in case people want to locate the window by the name of the command they used to start the application. But just "Spotify TUI" is probably nicer and maybe locating the window by spt is not enough to justify messing with it?

Thanks!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,7 +22,9 @@ use crossterm::{
   event::{DisableMouseCapture, EnableMouseCapture},
   execute,
   style::Print,
-  terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+  terminal::{
+    disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen, SetTitle,
+  },
   ExecutableCommand,
 };
 use network::{get_spotify, IoEvent, Network};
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -268,7 +268,12 @@ async fn start_ui(user_config: UserConfig, app: &Arc<Mutex<App>>) -> Result<()>
   execute!(stdout, EnterAlternateScreen, EnableMouseCapture)?;
   enable_raw_mode()?;
 
-  let backend = CrosstermBackend::new(stdout);
+  let mut backend = CrosstermBackend::new(stdout);
+
+  if user_config.behavior.set_window_title {  
+    backend.execute(SetTitle("spt - Spotify TUI"))?;
```
Rigellute, line 274: Very cool!

Rigellute (approved):
Nice addition. I think placing it in the `Behavior` struct makes sense 👍.

Thanks again 👏

Rigellute:
I think I've fixed these clippy errors in master. But would you mind rebasing your branch?

diegov:
> I think I've fixed these clippy errors in master. @diegov But would you mind rebasing your branch?

I've rebased master and fixed the rustfmt issues, clippy and rustfmt are passing locally now.

Rigellute:
Thanks @diegov and sorry for the delay in merging this.

Rigellute:
@all-contributors please add @diegov for code

Issue #868, opened 2021-08-25
Show `album_type` in Album search panes


jackson15j:
The Album search panes current show: `<album> - <artist>`, but there is no way to distinguish an `album` vs `single` (like the Spotify WebUI/Client).

Expect something like: `<album> - <artist> (<album_type>)`.

PR #869, opened 2021-08-25
Feature/show album type in search panes

jackson15j:
The Album search panes only showed: `<album> - <artist>`, so either required a User to know which item was an album/single/etc or select each item to see if the track list looks like an album/single.

This PR updates the Album search panes to the format: `<album> - <artist> (<album_type>)`, so that it is easier to find an `album` vs `single`. 

Fixes: #868

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -470,9 +470,10 @@ where
             }
           }
           album_artist.push_str(&format!(
-            "{} - {}",
+            "{} - {} ({})",
             item.name.to_owned(),
-            create_artist_string(&item.artists)
+            create_artist_string(&item.artists),
+            item.album_type.as_deref().unwrap()
           ));
           album_artist
         })
@@ -1237,9 +1238,10 @@ where
           }
         }
         album_artist.push_str(&format!(
-          "{} - {}",
+          "{} - {} ({})",
           item.name.to_owned(),
-          create_artist_string(&item.artists)
+          create_artist_string(&item.artists),
+          item.album_type.as_deref().unwrap()
         ));
         album_artist
       })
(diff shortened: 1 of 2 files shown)


src/ui/mod.rs
```diff
@@ -470,9 +470,10 @@ where
             }
           }
           album_artist.push_str(&format!(
-            "{} - {}",
+            "{} - {} ({})",
             item.name.to_owned(),
-            create_artist_string(&item.artists)
+            create_artist_string(&item.artists),
+            item.album_type.as_deref().unwrap()
```
sputnick1124, line 476: I don't think it's likely that an artist album has no type (I certainly haven't been able to find any), but I still don't love the naked `unwrap`. Can we change instead to something like `unwrap_or("unknown")`

src/ui/mod.rs
```diff
@@ -1237,9 +1238,10 @@ where
           }
         }
         album_artist.push_str(&format!(
-          "{} - {}",
+          "{} - {} ({})",
           item.name.to_owned(),
-          create_artist_string(&item.artists)
+          create_artist_string(&item.artists),
+          item.album_type.as_deref().unwrap()
```
sputnick1124, line 1244: Same here. Prefer `unwrap_or`

sputnick1124 (changes requested):
I think this is great!

src/ui/mod.rs
jackson15j: I'm happy with that. Pushed change in commit: 75f107f. Marking as Resolved.

src/ui/mod.rs
jackson15j: Same fix in commit: 75f107f. Marking as Resolved.

sputnick1124 (approved):

Rigellute (approved):
Sorry for the delay in reviewing this. Thank you, this is a nice addition 👍

PR #907, opened 2021-10-07
feat(config): check for and use XDG_CONFIG_HOME

hxlcyxn:
This checks if `XDG_CONFIG_HOME` is set and follows that, but defaults to `.config` if unset.

Because I personally use a non-default home directory layout, I was never able to use spotify-tui as it didn't respect my config directory, but I finally got around to "fix" it myself.


jasperspahl (approved):
Looks good

PR #910, opened 2021-10-24
Fix confirmation dialog handling on playlist delete.

### LINKED ISSUE #884: "D" to delete playlist does not work
Pressing `shift+D` according to the keybindings shown in the `?` menu does not successfully delete a playlist.

To reproduce:

1. Start spt
2. Enter the `Playlists` section
3. Select a playlist
4. Press shift+D to delete the selected playlist
5. Observe that nothing happens

I'm running spt 0.25.0 on Arch Linux with i3 wm, tried in both tmux and in a straight xfce4-terminal.

majabojarska:
Resolves #884. 

### Observations
I might not know the architecture of this application very well, but I believe I've narrowed down the root cause and drafted a quick solution. As of commit [ecddb5e](https://github.com/Rigellute/spotify-tui/commit/ecddb5eb4fb441bf46075d44cf12cd2294a251c7), the `push_navigation_stack` method of the `App` class does not actually push any element to the navigation stack if the next route's id is the same as last route's id:
```rust
// app.rs
pub fn push_navigation_stack(&mut self, next_route_id: RouteId, next_active_block: ActiveBlock) {
    if !self
      .navigation_stack
      .last()
      .map(|last_route| last_route.id == next_route_id)
      .unwrap_or(false)
    { // Enter branch only if the last route's id is different from the next route's id.
      // Push to stack...
    }
  }
```
That's the case when the uppercase `D` character is entered, when a playlist is selected - the last route's id is the same as the next route's id (`RouteId::Home`):
```rust
// playlist.rs
Key::Char('D') => {
      if let (Some(playlists), Some(selected_index)) = (&app.playlists, app.selected_playlist_index)
      {
        let selected_playlist = &playlists.items[selected_index].name;
        app.dialog = Some(selected_playlist.clone());
        app.confirm = false;

        let route = app.get_current_route().id.clone(); // Pass current route's id as the next route's id
        app.push_navigation_stack(route, ActiveBlock::Dialog(DialogContext::PlaylistWindow));
      }
    }
```
As a result, no new element is pushed to the navigation stack, hence the confirmation dialog window does not appear, and the playlist cannot be deleted.

### Solution
My solution is to add a `RouteId::Dialog`, which is specific to confirmation dialog windows. When a dialog is to be opened (e.g. when deleting a playlist), a new route (different than the current one) is pushed onto the navigation stack. Then, the logic responsible for drawing and handling user input in the confirmation dialog gets executed and works just fine.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -768,13 +768,8 @@ impl App {
       }
       None => {
         if let Some(saved_artists) = &self.library.saved_artists.clone().get_results(None) {
-          match saved_artists.items.last() {
-            Some(last_artist) => {
-              self.dispatch(IoEvent::GetFollowedArtists(Some(last_artist.id.clone())));
-            }
-            None => {
-              return;
-            }
+          if let Some(last_artist) = saved_artists.items.last() {
+            self.dispatch(IoEvent::GetFollowedArtists(Some(last_artist.id.clone())));
           }
         }
       }
--- a/src/app.rs
+++ b/src/app.rs
@@ -158,6 +158,7 @@ pub enum RouteId {
   Podcasts,
   PodcastEpisodes,
   Recommendations,
+  Dialog,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


Rigellute:
Excellent write up and excellent solution. Thank you!

Rigellute:
@majabojarska looks like there is a clippy error, are you able to fix that please?

majabojarska:
Fixed :wink:

Rigellute:
Great work @majabojarska, thanks again.

PR #926, opened 2021-12-20
feat: allow offset results of `spt list`

zigmars:


danbulant:
As a side note, something similar would be appreciated for playlists as well (yes I have more than 50 playlists saved).

Issue #984, opened 2022-05-05
the duration disappears with the progress bar


dmaglio:
![image](https://user-images.githubusercontent.com/446573/166905345-4042dc42-7784-4c95-9ef4-53fd98e17de8.png)

as you can see from the image, when the progress bar reaches and exceeds the duration, it disappears because it is the same color as the bar. can't change color when background color changes?

ghost:
Work around I have is change the text color to something different. the config option is 'playbar_progress_text'

PR #986, opened 2022-05-27
Update progress label to be visible if progress bar is same color

rnewton5:
This PR is in response to https://github.com/Rigellute/spotify-tui/issues/984.

It updates how the song_progress's label is set, to ensure the duration can be seen even if the text color matches the playbar's foreground color. If the label's color has been set to one that is different from the playbar's foreground color, this change will have no effect.

image

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -1040,18 +1040,26 @@ where
       } else {
         Modifier::empty()
       };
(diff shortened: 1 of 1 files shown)


Rigellute:
Thank you for this solution. I am using this in an experimental branch and will merge into main soon.

Issue #1000, opened 2022-06-30
Project Update


Rigellute:
I sincerely apologise for my lack of maintenance, I do love this project but I have not been able to dedicate any time to it - life/other commitments have been getting in the way! I hope to be able to make more space for development in the future.

A serious amount of work is needed to upgrade the underlying spotify library. What's more, the current architecture has grown in complexity and needs a rethink. I had spent time exploring this refactor, but ran out of steam trying to reimplement existing features. 

Just know that I do intend to return to it!

Rigellute:
It's rather fitting that this is issue #1000 😁

sewnie:
### **LETS FUCKING GOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO**

elijahmg:
@Rigellute Thx for your work. I faced some issues because of underlying library and since I needed only couple of feature, I've started to build my own pure cli tool. It's very mature. 
Here it is https://github.com/elijahmg/spotify-rust-cli
Thank you for inspiration

nopeless:
@Rigellute I feel like there are just too many bugs for this project to be viable for every day use and way many edge cases that cause discomfort

I really do like the concept of this project but for now, I feel like someone has to rewrite the codebase with better underlying ui engine

Oppen:
Hi @Rigellute. Thanks for all the work you put on this project. I'd like to say I love it but I got too late to the party to try it :( I'm sure I would tho :heart:
I was thinking, is there anything we can do to help keep the project alive? I thought maybe we could create a GitHub org so some more people can review and merge PRs. I see a lot of people has enough interest to contribute fixes, so what's missing the most is a few people in charge of the follow up.
Of course, my suggestion only applies if you agree, and setting you as owner (you deserve it), just with some other people helping to keep it afloat.

aromeronavia:
Yes, we are quite a few people interested in contributing to this <3

sewnie:
Me waiting for @Rigellute to have someone takeover the project or have a organization 💀 

Honestly, can someone here responsible enough make a fork of this project that takes PRs and accepts contributions, and fixes to this project? Many parts of it is broken (1000+ issues relating to broken)

sewnie:
i have now created a fork of this repository that lives at https://github.com/spotify-tui/spotify-tui; with some PRs merged and a cargo.lock update, but i do not know Rust if at all so i have applied some that do work.

sewnie:
[...]

PR #1022, opened 2022-09-11
Fix analysis bar theme setting not working

Daxanius:
When customizing the theme for spotify-tui, I noticed I couldn't customize the analysis bar colors.

At first I thought I was doing something wrong, but looking into the source, it looks like you forgot to properly add the setting to the config. This is what I've fixed in this quick fix.

### DIFF
--- a/src/user_config.rs
+++ b/src/user_config.rs
@@ -13,6 +13,8 @@ const APP_CONFIG_DIR: &str = "spotify-tui";
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct UserTheme {
+  pub analysis_bar: Option<String>,
+  pub analysis_bar_text: Option<String>,
   pub active: Option<String>,
   pub banner: Option<String>,
   pub error_border: Option<String>,
@@ -381,6 +383,8 @@ impl UserConfig {
       };
     }
 
+    to_theme_item!(analysis_bar);
+    to_theme_item!(analysis_bar_text);
     to_theme_item!(active);
     to_theme_item!(banner);
     to_theme_item!(error_border);


emilHof:
I noticed the same thing! The changes you made are the ones that worked for me locally too! Hope this gets merged soon!

emilHof (approved):
These changes did not produce any test failures, compilation errors, or issues when I implemented them locally!

Daxanius:
The developer of this project seems to be inactive unfortunately. The `rspotify` crate (and many others) has also been updated in the meantime. 

On top of that, this project still uses OpenSSL `1.1`, which is now known to have a major security vulnerability. 

I've looked into updating these dependencies myself, but so much seems to have changed that you're looking at potentially rewriting the entire project, unfortunately.

emilHof:
yes, that does kind of go with the current narrative of this app..  they did mention that they were currently too busy to do much of the maintenance work in [this](https://github.com/Rigellute/spotify-tui/issues/1000) issue. 

i did have some trouble with getting OpenSSL to work as well, bu was not even aware of the security issues it poses! 

At what point is it appropriate to start a fork if a maintainer is not so active? i am quite new to OSS, so there is a lot of etiquette that I am not aware of..

Daxanius:
The appropriate time would be whenever you feel like it. It's FOSS, you do what you want with it.

If you're not satisfied with how something is you make a fork and you fix it.

That said, I honestly think it's better to start something like this from scratch, as the issue you linked mentions it desperately needs an infrastructure rewrite.

I recently saw that someone was already making a new Spotify TUI from scratch, but it hasn't progressed much unfortunately.

I myself am looking to either keep this project going, or to make something myself. Either way both paths require work and time.

emilHof:
that makes sense. i would assume the answer to the question of what to do would also depend on possibly getting a group of people together who would be willing to maintain it. i may be wrong in this perception, but starting something new simply to have it fall to the wayside again seems somewhat of a waste...

how silly would it be to try and restructure the existing app and use some of the framework that has already been established, and see where it goes from there?

you are right, i am seeing this too now, the maintainer did mention that `spotify-tui` would need a major rewrite, yet i am also quite new to this and i am not sure how capable i would be of coming up with something better from scratch..

Daxanius:
Yeah...

Restructuring the current project would require you to understand it's code & architecture before making any major changes. This alone can take a long time. On top of that you'd have to throw away a lot of code and move and mess with stuff before having patched together something which works. 

I'd say this is silly to the max :)

emilHof:
You are making some very good arguments there.. From what I am putting together, both a rewrite and starting from scratch will take a lot of time. And time is valuable haha! I mean, as long as it somewhat works for now, I am content with it! :D

PR #1024, opened 2022-09-18
Showing release year on search result listings and artist pages

danielpietzsch:
I was missing a “Release Year” info on search result and artist pages, which would help me decide what release from an artist I am looking for.

When I forked this, I noticed, that @jackson15j recently had [already done something similar with album _types_](https://github.com/Rigellute/spotify-tui/pull/869) (which I also greatly appreciate). So I build my solution on top of _that_.

Hope you’ll find this worthwhile, too, @Rigellute.

On smaller screens and/or with long album titles, all this new info – especially this new “release year” addition – might not show, because it’s _appended_ to the release’s name (and hence will be cut off). But I still found this to be useful for a lot of artists and result pages.

### DIFF
--- a/src/ui/util.rs
+++ b/src/ui/util.rs
@@ -1,6 +1,7 @@
 use super::super::app::{ActiveBlock, App, ArtistBlock, SearchResultBlock};
 use crate::user_config::Theme;
 use rspotify::model::artist::SimplifiedArtist;
+use rspotify::model::album::SimplifiedAlbum;
 use tui::style::Style;
 
 pub const BASIC_VIEW_HEIGHT: u16 = 6;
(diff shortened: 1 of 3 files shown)


src/ui/util.rs
```diff
@@ -47,6 +48,14 @@ pub fn create_artist_string(artists: &[SimplifiedArtist]) -> String {
     .join(", ")
 }
 
+pub fn create_album_type_and_release_year_string(album_item: &SimplifiedAlbum) -> String {
+  // No matter the release_date_precision, release_date should always start with a four-digit string of the year (YYYY[-MM[-DD]]).
+  // Getting those first four letters/digits should always work.
+  let album_release_year = album_item.release_date.as_deref().unwrap().get(0..4).unwrap();
+
+  album_item.album_type.as_deref().unwrap_or("unknown").to_owned() + ", " + album_release_year
```
danielpietzsch, line 56: If the above code can be improved, I’m all ears!

PR #1041, opened 2022-11-09
Performs full dependency update

xarthna:
Does not change `Cargo.toml` but issues a `cargo update` for full dependency update.

The reason for this PR is due to Artix Linux rolling release distro (and other distros) moving to openssl-3.* and dropping openssl-1.1. Therefore, `spt` cannot find `libssl.so.1.1` at runtime and needs to be rebuilt against the newer ssl version. The main target package for upgrade is [rust-openssl v0.10.42](https://github.com/sfackler/rust-openssl/releases/tag/openssl-v0.10.42) which is not a top level dependency of the project. According to the rust-openssl project the new version v0.10.42 [supports](https://docs.rs/openssl/latest/openssl/) openssl 1.0.1 through 3.x.x. Therefore, I believe this update should be backwards compatible with openssl-1.1 installations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.14.1"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


VoidAny:
Yep. Running `cargo update` fixed the problem for me.

Breathof (approved):

PR #1087, opened 2023-06-11
added more infos

idkwuu:


README.md
```diff
@@ -1,5 +1,8 @@
 # Spotify TUI
 
+> **Note**
+> make sure you have **Spotify Premium**.
```
sewnie: This is already listed down below as:
> If you want to play tracks, Spotify requires that you have a Premium account.

PR #1095, opened 2023-08-18
Remove unused mut

Sanceilaks:


src/app.rs
```diff
@@ -663,7 +663,7 @@ impl App {
     active_block: Option<ActiveBlock>,
     hovered_block: Option<ActiveBlock>,
   ) {
-    let mut current_route = self.get_current_route_mut();
+    let current_route = self.get_current_route_mut();
```
BinaryArchaism, line 666: Hello! 
Why? You won't assign `active_block` on 668 line

PR #1139, opened 2024-11-14
docs: add new section explaining how to change the used account

Letder40:
I created a new branch called docs, where I added the changes to the documentation, I included how an user can change the Spotify's account used in spotify-tui.


sewnie:
spotify-tui is abandonware, it doesn't work, use spotify_player.

Letder40:
ooh thank you, but it's working fine for me, I will checkout spotify_player.
Why it shouldn't work?

PR #1151, opened 2025-04-16
API callback listener use 127.0.0.1 instead of localhost

### LINKED ISSUE #1150: Spotify auth failed
Hi, Spotify no longer lets you use localhost as part of the redirect_uri which breaks the auth, i tried using http://127.0.0.1:8888 instead which sadly doesn't work.

https://developer.spotify.com/documentation/web-api/concepts/redirect_uri

tobraha:
R.I.P. spotify-tui :sob: 

Spotify client apps now require HTTPS except for localhost callback listeners, but it only recognizes `127.0.0.1` and not `localhost`.

This updates the instructions and callback reference.

Resolves #1150

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ But here they are again:
 1. Click `Create an app`
     - You now can see your `Client ID` and `Client Secret`
 1. Now click `Edit Settings`
-1. Add `http://localhost:8888/callback` to the Redirect URIs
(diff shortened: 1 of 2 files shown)


Roelof-hue:
If you know good alternatives let me know!

sewnie:
https://github.com/aome510/spotify-player

PR #1163, opened 2025-11-02
Change redirect URI from localhost to 127.0.0.1

AntoneBrewer:
Updated the redirect URI to use 127.0.0.1 instead of localhost for better compatibility and consistency with IP-based configurations.


AntoneBrewer:
Updated the redirect URI to use 127.0.0.1 instead of localhost for better compatibility and consistency with IP-based configurations.

AntoneBrewer (commented):
Replaced old uri with new one

mhsFlairs (approved):

